PR #855, opened 2024-11-16
Improvements to Debug Processing:Core

Stefterv:
We need a way to debug / run sketches independently of the Processing IDE. Later I would like to add the ability to debug sketches through the PDE, e.g. Run a sketch in the PDE and IntelliJ will attach the debugger to the ran sketch


Stefterv:
One downside of this approach, the basic sketch gets included in processing:core

PR #869, opened 2024-12-06
Sign macos prebuilds

Stefterv:
Signing the macOS pre-builds so we can test pre-releases on our own machines. Unfortunately signing the application is not enough and we need to notarise it if we want it to be able to be run. From here we have two options:
- Notarise pre-builds anyway, I dislike this option as it allows for unverified software to be distributed by under the Processing Foundation's name
- Instruct people to use the following hack: `xattr -d com.apple.quarantine ./Processing.app` this is a more advanced use case but the people who need to check beta software could be expected to jump through this hoop...

If anyone has any input let me know

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -66,8 +66,16 @@ jobs:
           architecture: ${{ matrix.arch }}
       - name: Setup Ant
         uses: cedx/setup-ant@v3
+      - name: Install Certificates for Code Signing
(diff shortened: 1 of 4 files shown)


SableRaf:
Thanks, @Stefterv! I think the `xattr -d com.apple.quarantine` workaround is the way to go. We could include this information in a dedicated `INSTRUCTIONS_FOR_TESTING` file specifically for test builds. What do you think?

PR #899, opened 2025-01-03
Fix: set stroke and fill properly in PFont.getShape

Junology:
Currently, `PFont.getShape` returns an instance of `PShape` which renders nothing.
This is because fill and storke are not properly set in the construction of the shape.
The commit fixes the issue by calling `PShape.noStroke()` and `PShape.fill(0)` at the begining of the construction.

Test code:
```processing
import java.awt.Font;
import java.util.stream.IntStream;

PFont symbols;
PShape[] symbolShapes;

void setup() {
  size(512, 512);
  
  symbols = createFont("winjs-symbols.ttf", 32);
  Font font = (Font) symbols.getNative();
  final int[] codes = IntStream.range(0, 0x10000).filter(font::canDisplay).toArray();
  symbolShapes = new PShape[codes.length];
  for(int i = 0; i < codes.length; ++i) {
    symbolShapes[i] = symbols.getShape((char) codes[i]);
    symbolShapes[i].width=32;
    symbolShapes[i].height=32;
  }
}

void draw() {
  background(#AABBCC);
  for(int i = 0; i < symbolShapes.length; ++i) {
    int x = i % 16;
    int y = i / 16;
    symbolShapes[i].setFill(#0000FF);
    shape(symbolShapes[i], 32 * x + 4, 32 * y + 36, 24, 24);
  }
}
```
`winjs-symbols.ttf` can be downladed [here](https://github.com/microsoft/fonts/raw/1259517be73038f430d6e52fd130dd714bef6804/Symbols/winjs-symbols.ttf).

### DIFF
--- a/core/src/processing/core/PFont.java
+++ b/core/src/processing/core/PFont.java
@@ -740,7 +740,7 @@ public class PFont implements PConstants {
     // six element array received from the Java2D path iterator
     float[] iterPoints = new float[6];
     // array passed to createGlyphVector
-    char[] textArray = new char[] { ch };
+    char[] textArray = { ch };
 
     //Graphics2D graphics = (Graphics2D) this.getGraphics();
     //FontRenderContext frc = graphics.getFontRenderContext();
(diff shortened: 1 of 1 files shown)


Junology:
Another concern is the "origin" of the shape generated by `PFont.getShape`.
Namely, since the coordinate (0,0) is on the base line of the character, most of the part of the shape is *above* the origin, which may be inconsistent with the other shapes which are usually rendered *below* the origin.
However, the "height" of the character shapes completely depends on their use cases, so "on the base line" might be only a sane choice.
IMHO, `PShape` can have a functionality to set its origin after the creation (like we can for `width` and `height`); cf #788 .

Junology:
I am really sorry that I mistakenly close the PR.
I didn't notice the PR was made based on `main` branch of my fork, so I heavily messed it.
I eventually remake this PR.

PR #902, opened 2025-01-08
Make PShapeOpenGL.setFill(int,int) always update tessellated vertices

### LINKED ISSUE #900: PShape setFill(indx, color) not working in Processing 4

OpenGL

### Processing version

4.3

### Operating system

MacOSX

### Steps to reproduce this

"1. The source code below demonstrates the problem.  The PShape should change color when the mouse is clicked.

2. The source code runs correctly in versions 2.2.1 (macos) and 3.5.4 (reported by another forum member running Windows) indicating a runtime code change prior to version 4.

3. There is a similar report from a year ago: https://github.com/processing/processing4/issues/677"

4. We would like to change the vertex colors of the original PShape object without having to create a second object.

### snippet

```processing

PShape t;

void setup() {
  size(400, 400, P2D);
  surface.setTitle("Should change color when mouse clicked.");
  t = createShape();
  t.beginShape();
  t.vertex(200, 100);
  t.vertex(100, 300);
  t.vertex(300, 300);
  t.endShape(CLOSE);
  for (int i = 0; i < t.getVertexCount(); i++) {
    t.setFill(i, color(random(255), random(255), random(255)));
  }
}

void draw() {
  background(209);
  shape(t);
}

void mousePressed() {
  for (int i = 0; i < t.getVertexCount(); i++) {
    t.getVertex(i);
    t.setFill(i, color(random(255), random(255), random(255)));
  }
}
```

### Additional context

_No response_

Junology:
Fixes #900 (and #677 too).

`PShapeOpenGL` provides "Tessellation update mode" `beginTessellation()` and `endTessellation()`.
The current implementation of `PShapeOpenGL.setFill(int,int)` does the following:
- if it is in "Tessellation update mode" (`root.tessUpdate == true`), it updates the color associated with tessellated vertices;
- otherwise, it updates the "raw" data `InGeometry` and set "tessellation required" flag with `markForTessellation()`.

The problem is that once the shape is tessellated, the tessellation doesn't seem to be performed again even after markForTessellation().
Another concern is that the behavior described above is inconsistent with `PShape.OpenGL.setFill(int)` overload, which does not care about "Tessellation update mode" and updates tessellated vertices anyway.

This PR makes `PShapeOpenGL.setFill(int,int)` behaves like the other overloads.
As a result, it also ignores "Tessellation update mode".
I am not 100% sure it is safe, since `beginTessellation()` and `endTessellation()` look doing a lot of stuff.
In particular, I think this change should be tested for the cases where tessellation creates new vertices.
Since I do not know the tessellation process in detail, such test cases are really appreciated.

### DIFF
--- a/core/src/processing/opengl/PShapeOpenGL.java
+++ b/core/src/processing/opengl/PShapeOpenGL.java
@@ -2005,12 +2005,12 @@ public class PShapeOpenGL extends PShape {
     }
 
     if (image == null) {
-      if (root.tessUpdate) {
+      inGeo.colors[index] = PGL.javaToNativeARGB(fill);
+      if (shapeCreated && tessellated && hasPolys) {
         int tessIdx = firstPolyVertex + index;
         tessGeo.polyColors[tessIdx] = PGL.javaToNativeARGB(fill);
         root.setModifiedPolyColors(tessIdx, tessIdx);
       } else {
-        inGeo.colors[index] = PGL.javaToNativeARGB(fill);
         markForTessellation();
       }
     }
@@ -2101,12 +2101,12 @@ public class PShapeOpenGL extends PShape {
     }
 
     if (image != null) {
-      if (root.tessUpdate) {
+      inGeo.colors[index] = PGL.javaToNativeARGB(tint);
+      if (shapeCreated && tessellated && hasPolys) {
         int tessIdx = firstPolyVertex + index;
         tessGeo.polyColors[tessIdx] = PGL.javaToNativeARGB(tint);
         root.setModifiedPolyColors(tessIdx, tessIdx);
       } else {
-        inGeo.colors[index] = PGL.javaToNativeARGB(tint);
         markForTessellation();
       }
     }


hx2A:
Thank you, @Junology , for this PR!

Does `setStroke()` need a similar code change? We should probably review the set methods in this class to ensure that where appropriate there is some consistency between their implementations.

Junology:
If this change makes sense, then yes, `setStroke()` also needs a similar change.
Thank you for pointing it out; I will make a commit on that later.
But, before anything, I'd like to make sure it is really safe; the tessellation process is too complicated to understand all the details.

Exploring `git log`, I found that the "tessellation update mode" `beginTessUpdate`/`endTessUpdate` (later renamed to the current ones) were introduced in the commit 6595427e117b31e685fe3e53477410854f88c0bb.
Does someone remember the logic behind the mode at that point?
In particular, what types of operations are supposed to be enclosed with those functions?
Is it really safe to update colors outside of the mode, and how about the other attributes, especially when tessellation creates some new vertices?

Junology:
I realized the change done in this PR is ***UNSAFE***.
Indeed, the modified version of `setFill(int,int)` can cause `ArrayIndexOutOfBoundsException` if it is used in "tessellation update mode".

Here is the code I tested:
```processing
PShape[] t;

void setup() {
  size(300, 200, P2D);
  t = new PShape[2];
  for (int i = 0; i < t.length; ++i) {
    t[i] = createShape();
    t[i].beginShape();
    t[i].fill(#FFFFFFFF);
    t[i].vertex(0, 0);
    t[i].vertex(0, 100);
    t[i].vertex(100, 0);
    t[i].vertex(100, 100);
    t[i].endShape(CLOSE);
  }
}

void draw() {
  background(0);
  float offset = (float) (width - 100 * t.length) / (t.length + 1);
  for (int i = 0; i < t.length; ++i) {
    shape(t[i], offset + (100 + offset) * i, 50);
  }
}

void mousePressed() {
  // Reset fill colors in different ways
  t[0].setFill(#FFFF0000);

  // setFill(int,int) in "tessellation update mode" can cause ArrayIndexOutOfBoundsException
  // t[1].beginTessellation();
  for (int i = 0; i < t[1].getVertexCount(); i++) {
    t[1].setFill(i, #FFFF0000);
  }
  // t[1].endTessellation();
}
```
First notice that the above program does not work as expected; the color of the shape `t[1]` changes only partially though the `for`-loop seems traversing all the vertices.
Second, uncommenting the lines of `beginTessellation()`/`endTessellation()` causes `ArrayIndexOutOfBoundsException`.

The source of the exception lies in how the function `getVertexCount()` behaves; it usually counts the vertices in the raw shape (i.e. `inGeo`), but in "tessellation update mode", it looks at the tessellated shape `tessGeo`.
Hence, if there are new vertices created in tessellation, the returned value can exceed the size of the vertex array of `inGeo`, and that is why the exception.

I decided to close the PR, thinking it is difficult to fix the problem mentioned above by refactoring `setFill(int,int)` itself; for a vertex in the raw shape given, there is no way to determine the corresponding vertex in the tessellated shape.

If one is sure what happens in tessellation, then one can use the function in "tessellation update mode" anyway.
On the other hand, if one wants to use it without knowing tessellation, then fixing #900 is still necessary.
Making `markForTessellation()` really cause re-tessellation seems to be the only way to go.

PR #909, opened 2025-01-10
Update Contributions url

Stefterv:
Changes the download url for Processing contributions. This is part of a larger transition strategy moving to a new contributions workflow: https://github.com/processing/processing-contributions


Stefterv:
Thinking about it, since the url is arbitrary, maybe it makes more sense to keep it original as that will make it more likely that we will be backwards compatible in the future.

PR #917, opened 2025-01-15
Build libDifferent CI

### LINKED ISSUE #913: Creating a CI/CD stage for `ThinkDifferent` class
As shown by @hx2A's Pull Request #912 we need a systematic way to compile the `libDifferent` binaries. Initially I wanted to add it to Gradle but thinking about it a bit more, I think we should just build it as a standalone Github Action. 

It would be triggered by any of the code in `core/different` changing. Then it would compile the new binaries and create a pull request back into the repo.

Some considerations for doing this:
- I'd like the code to be compiled in a more transparent manner
- I'd like to make sure the methodology of compiling the binaries is somehow captured in code and not just documentation
- Security wise, I think it is more secure to have the binary compiled by a trusted source rather than a contributor that decides to create changes. 

This way contributors can contribute changes to the `libDifferent` but will ultimately always be overriding by a own compiled binary.

Stefterv:
Closes #913 

This GitHub Action will:

- Trigger when changes are pushed to either core/different/different.m or core/different/Makefile
- Create a pull request with the new binaries
   - An example of this can be found here: https://github.com/Stefterv/processing4/pull/4

### DIFF
--- a/.github/workflows/different.yml
+++ b/.github/workflows/different.yml
@@ -1,0 +1,40 @@
+name: Build libDifferent
+
+on:
+  push:
+    branches:
+      - main
+    paths:
+      - 'core/different/different.m'
+      - 'core/different/Makefile'
+
+jobs:
+  build-and-pr:
+    runs-on: macos-latest
+    permissions:
+      contents: write
+      pull-requests: write
+
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Run make in core/different
+        run: |
+          cd core/different
+          make
+
+      - name: Create Pull Request
+        uses: peter-evans/create-pull-request@v5
+        with:
+          commit-message: 'Update generated files from make'
+          title: 'Update generated files from make'
+          body: |
+            This PR was automatically created after changes to:
+            - core/different/different.m
+            - core/different/Makefile
+            
+            The `make` command was run and these are the resulting changes.
+          branch: auto-different-updates
+          base: main


hx2A:
I'm not the best person to review github actions, but from what I know about them, this looks good to me.

One more change I'd suggest is that we add the compiled library file to `.gitignore` to discourage anyone from including them in a commit or future PR. I realize this github action would overwrite library files created by PR submitters and adding it to `.gitignore` doesn't prevent anyone from overriding the ignore directive. I still think it would be a good idea to include it there though.

Stefterv:
Yeah I thought about that but then the ci for the pull request previews wouldn't work. 

As example, you submitted a change and I would like to test that change, then I would need a build with the new binary in Processing to make sure it works. I think the only way would be to allow the binary to be part of the PR

hx2A:
> Yeah I thought about that but then the ci for the pull request previews wouldn't work.
> 
> As example, you submitted a change and I would like to test that change, then I would need a build with the new binary in Processing to make sure it works. I think the only way would be to allow the binary to be part of the PR

Oh, I see. You need the compiled binary to be in the PR so the CI testing will work. There's no way for the CI stuff to compile it itself, before running tests?

Stefterv:
Good you ask, I just realized only macOS can export for macOS so maybe it can just be part of the build chain.

I was under the impression that libDifferent needed to be included on other platforms too

In any case `core` is compiled and posted on Linux and I'd prefer not to change that.

hx2A:
OK, good we talked this through and understand how everything will work moving forward.

hx2A (approved):
Looks good! This will compile the native thinkdifferent library, needed for macOS computers

Issue #911, opened 2025-01-11
Disable/control blank lines in the beginning of the console


Junology:
### Most relevant area for this enhancement?

Accessibility

### Feature to enhance

Processing prints 10 blank lines when it runs PDE

### Desired enhancement

In Linux, Processing IDE is launched by `java` command with `&` in the tail, so the blank lines just mess up the terminal; e.g.
```
user@processing-4.3$ ./processing
user@processing-4.3$ 









echo "This is the next command"
This is the next command
user@processing-4.3$ 
```
I just want to disable the feature, but regarding backward compatibility, let me suggest three options:

1. Just stop printing blank lines. The simplest solution, but this can break backward compatibility.
2. Add configurable option to `defaults.txt` or `preferences.txt` like `console.blanklines` with default value `10` which controls the number of blank lines. A user can set it to `0` to disable the printing. Also, it would not affect existing projects unless they depend on `defaults.txt` itself.
3. Do nothing. Maximal backward compatibility.

### Potential benefits

The enhancement may also benefit GUI users since the blank lines are also printed to IDE's "console" area.
For those users, the blank lines make little sense.
![editor-blankline](https://github.com/user-attachments/assets/d6158670-35f1-492e-9b4c-6f76b2a4b09e)

### Possible drawbacks

According to the comment in the source code, the blank lines are for Windows console (see below).
However, the line was added in Processing 2 era, so things may have changed.
Backward compatibility can also be a problem.

### Additional context

The blank lines come from `Editor.prepareRun()` method in the line 2297 in `processing.app.ui.Editor`.
```java
  public void prepareRun() {
    internalCloseRunner();
    statusEmpty();

    // do this to advance/clear the terminal window / dos prompt / etc
    for (int i = 0; i < 10; i++) System.out.println();
```

### Would you like to work on the issue?

Once the direction becomes clear, I think I can do it by myself.

Stefterv:
Any idea why the terminal window needs to be cleared a bit in the first place?

Stefterv:
I'm also happy if it gets moved down a little bit:
https://github.com/processing/processing4/blob/5e1161c13733e2739126170701ab6006e26cbd66/app/src/processing/app/ui/Editor.java#L2297-L2302

Junology:
I have no idea why this clearance is needed in "the terminal windows / dos prompt / etc", at least in the current environment around Processing.
According to `git log`, these lines were introduced in 2011, but I don't know about the situation in that time since I am pretty new here.
[...]

PR #919, opened 2025-01-17
Add new preference option 'console.head_padding'

Junology:
The PR adds `console.head_padding` preference option to configure blank lines printed in terminal/console in each time a sketch is run.

The background is found in #911.

Making the change, I originally planned to only add an option for backward compatibility.
However, I finally decided to rewrite the code for "padding by blank lines" using `EditConsole.message` instead of `System.out.println` to stop printing them in the terminal.
The reasons why I think this is safe are as follows:
1. First of all, the change only affects the IDE (i.e. `processing`) and not command-line mode (i.e. `processing-java`).
Hence, it would not affect the final products of any projects using Processing either.
2. **Windows**: after investigation, I realized that the blank lines does not affect Windows terminals (i.e. command prompt or PowerShell) probably because of launch4j.
So, maybe it would not seriously break backward compatibility in Windows if the blank lines are removed.
3. **Linux**: `processing` is run as a background process, so any outputs can mess up the terminal.
Moreover, since the blank lines are likely the only outputs from IDE (NOTE: `print`/`println` does not print to terminal from IDE), so I think nobody rely on the behavior.

Unfortunately, I have not tested on MacOS and cannot since I don't have any machines.
It would be great if someone tests the change of behavior on MacOS and gives comments.

I also note that, in spite of my comments in #911, the head padding may be useful when `console.auto_clear` option is set to `false`.
For example, the following code prints `HelloHelloHello` without line breaks if you run it three times with `console.auto_clear = false` and `console.head_padding = 0` on your `preferences.txt`:
```processing
print("Hello");
```

I wonder if `console.auto_clear` and the new option `console.head_padding` are worth putting in the preference dialog.
In some situations, keeping console outputs may help debugging, but currently nobody knows the option.

### DIFF
--- a/app/src/processing/app/ui/Editor.java
+++ b/app/src/processing/app/ui/Editor.java
@@ -2293,19 +2293,9 @@ public abstract class Editor extends JFrame implements RunnerListener {
     internalCloseRunner();
     statusEmpty();
 
-    int headPadding;
-
-    try {
-      headPadding = Preferences.getInteger("console.head_padding");
-    } catch (NullPointerException e) {
-      // We need to handle this exception to take care of old versions of
-      // preference files, i.e., "defaults.txt" and "preferences.txt" which
-      // may not have the attribute 'console.head_padding'.
-      headPadding = 10;
-    }
-
     // Do this to advance/clear the terminal window / dos prompt / etc.
     // This may be useful especially when 'console.auto_clear' is false.
+    int headPadding = Preferences.getInteger("console.head_padding");
     for (int i = 0; i < headPadding; i++) console.message("\n", false);
 
     // clear the console on each run, unless the user doesn't want to
--- a/app/src/processing/app/ui/Editor.java
+++ b/app/src/processing/app/ui/Editor.java
@@ -2293,8 +2293,22 @@ public abstract class Editor extends JFrame implements RunnerListener {
     internalCloseRunner();
     statusEmpty();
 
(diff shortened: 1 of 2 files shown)


app/src/processing/app/ui/Editor.java
```diff
@@ -2293,8 +2293,20 @@ public void prepareRun() {
     internalCloseRunner();
     statusEmpty();
 
-    // do this to advance/clear the terminal window / dos prompt / etc
-    for (int i = 0; i < 10; i++) System.out.println();
+    int headPadding;
+
+    try {
+      headPadding = Preferences.getInteger("console.head_padding");
+    } catch (NullPointerException e) {
```
Stefterv: I don't see any other examples of the codebase wrapping the `getInteger` with a try catch so I don't think this is necessary

Stefterv (approved):
Thank you for your contribution @Junology! I'm happy to add it and thank you for documenting your work thoroughly!

If you have a bit more time, please check the try catch otherwise this is good to go!

app/src/processing/app/ui/Editor.java
Junology: You mean we don't need to be afraid of de-coupling of the codebase and `defaults.txt`, right? If so, I agree that `try`-`catch` doesn't make sense.

Stefterv (approved):

PR #921, opened 2025-01-17
Make the specification for the runner more specific

Stefterv:
Right now we have arm64 and arm32 runners and sometimes they get mixed up because the label is not specific enough causing builds to fail 

[skip ci]


SableRaf:
Thanks for tagging me for review! I might be missing something, but I don't see a distinction between `arm32` and `arm64` in this PR, only the added labels for `Linux` and `ARM`. Could you clarify how these changes address the issue?

Stefterv:
Yes so this is the lovely situation where ARM means arm32 and ARM64 means arm64

PR #927, opened 2025-01-17
Export to PDEZ button

Stefterv:
Screenshot 2025-01-17 at 19 55 10

Adds button to the File menu to package the current sketch into a PDEZ file


Stefterv:
I just happened to test it with a sketch that was also exported and it bundled the export application in the PDEZ, do we want to prevent this?

SableRaf:
Maybe we should make it a dialog? Something like 

> _"Your sketch folder contains exported applications (executable versions of your sketch, built for a specific operating system). Do you want to include them in the .pdez bundle? Including them will make the file larger."_

SableRaf (approved):

Issue #918, opened 2025-01-16
NullPointerException thrown in `PApplet.dequeueWindowEvents()`


hx2A:
I had several running Sketches and was moving the windows around when this exception was thrown:

```txt
java.lang.NullPointerException: Cannot invoke "java.lang.Integer.intValue()" because the return value of "java.util.Map.remove(Object)" is null
        at processing.core.PApplet.dequeueWindowEvents(PApplet.java:9740)                                        
        at processing.core.PApplet.handleDraw(PApplet.java:2085)
        at processing.opengl.PSurfaceJOGL$DrawListener.display(PSurfaceJOGL.java:840)                             
        at jogamp.opengl.GLDrawableHelper.displayImpl(GLDrawableHelper.java:692)
        at jogamp.opengl.GLDrawableHelper.display(GLDrawableHelper.java:674)
        at jogamp.opengl.GLAutoDrawableBase$2.run(GLAutoDrawableBase.java:443)                                    
        at jogamp.opengl.GLDrawableHelper.invokeGLImpl(GLDrawableHelper.java:1293)
        at jogamp.opengl.GLDrawableHelper.invokeGL(GLDrawableHelper.java:1147)
        at com.jogamp.newt.opengl.GLWindow.display(GLWindow.java:782)
        at com.jogamp.opengl.util.AWTAnimatorImpl.display(AWTAnimatorImpl.java:81)
        at com.jogamp.opengl.util.AnimatorBase.display(AnimatorBase.java:453)
        at com.jogamp.opengl.util.FPSAnimator$MainTask.run(FPSAnimator.java:178)
        at java.base/java.util.TimerThread.mainLoop(Timer.java:566) 
        at java.base/java.util.TimerThread.run(Timer.java:516)
```

I see what the problem is. In the code there is this:

```java
  /**
   * Internal use only: called by Surface objects to queue a position
   * event to call windowPositioned() when it's safe, which is after
   * the beginDraw() call and before the draw(). Note that this is
   * only the notification that the window is in a new position.
   */
  public void postWindowMoved(int newX, int newY) {
    if (external && !fullScreen) {
      frameMoved(newX, newY);
    }

    windowEventQueue.put("x", newX);
    windowEventQueue.put("y", newY);
  }


  /** Called when the window is moved */
  public void windowMoved() {  }


  private void dequeueWindowEvents() {
    if (windowEventQueue.containsKey("x")) {
      windowX = windowEventQueue.remove("x");
[...]

PR #930, opened 2025-01-19
Fix for NullPointerException thrown in `PApplet.dequeueWindowEvents()`

hx2A:
Here is a fix for #918.

When the window is moved or resized, the new location or size values will be stored as a pair in the `windowEventQueue` map. This ensures that there are no race conditions if one value has been added but the other has not.

The exception I saw when I opened #918 is probably very rare and would be hard to reproduce. Nevertheless, we can review the code change and verify that the necessary functionality still works. You can test window move and resize events with the following code:

```java
void setup() {
  size(200, 200, JAVA2D);
  windowResizable(true);
  windowMove(100, 100);
}

void draw() {
}

void windowResized() {
  println("new size", width, height);
}

void windowMoved() {
  println("new location", windowX, windowY);
}

void keyPressed() {
  windowResize(200, 200);
  windowMove(100, 100);
}
```

It also works with the `P2D` renderer.

### DIFF
--- a/core/src/processing/core/PApplet.java
+++ b/core/src/processing/core/PApplet.java
@@ -9663,7 +9663,7 @@ public class PApplet implements PConstants {
 
   // WINDOW METHODS
 
-  Map<String, Integer> windowEventQueue = new ConcurrentHashMap<>();
+  Map<String, WindowEventValuePairs> windowEventQueue = new ConcurrentHashMap<>();
 
 
   public void windowTitle(String title) {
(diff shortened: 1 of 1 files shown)


Stefterv (approved):

Stefterv:
Hi @hx2A Thank you for the fix, I tested it on macOS arm64 and works just fine here too.

hx2A:
> Hi @hx2A Thank you for the fix, I tested it on macOS arm64 and works just fine here too.

You are welcome! Happy to help out here.

SableRaf (approved):

PR #943, opened 2025-02-05
Disable sketch folder deletion

Stefterv:
Solves https://github.com/processing/processing4/issues/794

When you delete the last tab in a sketch, Processing treats the entire folder containing the `.pde` file as the sketch folder. This means that if the `.pde` file is stored at the root of an important folder (like your Downloads or Desktop folder), **deleting it via Processing and confirming the prompt will delete everything in that folder.**

This causes confusion and has resulted in loss of data for some users. Show a warning message instead.

warning message that deleting the sketch folder is not supported


SableRaf (approved):
Looks good to me! 

Note: I added auto-generated translations but we could only really verify the French, Italian, and Dutch languages. Others were checked for obvious mistakes, but more detailed proof-reading will be necessary. That said, I think this is good for merging.

PR #955, opened 2025-03-07
Patch for keyPressed true with no keys pressed after specific sequence of ```SHIFT``` or ```CTRL```

### LINKED ISSUE #779: `keyPressed` is `true` with no keys pressed after specific sequence of `SHIFT` and any letter key
*Created by: keppy442b*

 

## Description

`keyPressed` is `true` with no keys actively pressed after doing a specific sequence of key presses with `SHIFT` and a letter key. It also occurs with `CTRL` and a letter key. This is more easily reproduced with P2D, but it does occur with the default renderer as well.

## Expected Behavior

`keyPressed` should be `false` when no keys are actively pressed.

## Steps to Reproduce

1. Press and hold a letter key
2. While still holding letter key, press and hold `SHIFT`
3. While still holding `SHIFT`, release the letter key
4. Release `SHIFT`

Pressing `CTRL` instead of `SHIFT` will also produce the behavior. It appears the only way to set `keyPressed` back to `false` is to press/release the specific letter key again. So if you did `SHIFT`+`a`, press and release `a` to set `keyPressed` back to the expected behavior. You can try to reproduce the behavior with this sketch:

```
color black = color(36,39,45);
color green = color(141,187,110);
void setup(){
  size(1600,900,P2D); 
}

void draw(){
  if (keyPressed){
    background(green);
  } else{
    background(black);
  }
}
```

## Your Environment

 

* Processing version: 4.3
* Operating System and OS version: Windows 10 Home 22H2 19045.3324

Rishab87:
The issue was that earlier we were using a hash for registering keys. For example, when someone pressed Ctrl and then A, the combination Ctrl+A got registered, and when they released Ctrl and then A, the Ctrl+A combination remained registered because it had a specific hash that was not released upon releasing both keys. That is why I removed the hash and am now directly registering the key.

However, I'm not sure why we were using a hash in the first place. Is there any specific reason? If yes, then I can change this approach.

Fixes #779

### DIFF
--- a/core/src/processing/core/PApplet.java
+++ b/core/src/processing/core/PApplet.java
@@ -2630,13 +2630,15 @@ public class PApplet implements PConstants {
 
     switch (event.getAction()) {
       case KeyEvent.PRESS -> {
-        Long hash = ((long) keyCode << Character.SIZE) | key;
+        Long hash = (long) keyCode;
         if (!pressedKeys.contains(hash)) pressedKeys.add(hash);
         keyPressed = true;
         keyPressed(keyEvent);
       }
       case KeyEvent.RELEASE -> {
-        pressedKeys.remove(((long) keyCode << Character.SIZE) | key);
+        Long hash = (long) keyCode;
+        pressedKeys.remove(hash);
+
         keyPressed = !pressedKeys.isEmpty();
         keyReleased(keyEvent);
       }


Stefterv:
Hi @Rishab87 the hash is combining the`keyCode` and `key` variables. I agree it's not very readable but I don't think switching back to only the keyCode is going to solve #779 properly. 

Please try to replicate the issue in your tests in #966 and then feel free to solve it in that PR too!

Thank you!

Rishab87:
@Stefterv I've already replicated it in tests and it was working fine with this approach although I'm not sure about if this approach can cause any further issues, I'll try some different approach with hash present and fix it in that PR, thank you for the suggestion.

PR #957, opened 2025-03-07
Move `java/preprocessor` to Gradle

Stefterv:
Changes the preprocessor from the ant build-system to gradle 

- [x] Fix the Maven release script: 

          FAILURE: Build failed with an exception.
                * What went wrong:
                A problem was found with the configuration of task ':java:preprocessor:sourcesJar' (type 'Jar').
                  - Gradle detected a problem with the following location: '/home/runner/work/processing4/processing4/java/preprocessor/build/generated-src/antlr/main'.
                    
                    Reason: Task ':java:preprocessor:sourcesJar' uses this output of task ':java:preprocessor:generateGrammarSource' without declaring an explicit or implicit dependency. This can lead to incorrect results being produced, depending on what order the tasks are executed.


Stefterv:
Is going to be part of the upcoming Gradle Runner

PR #958, opened 2025-03-07
Move `java/lsp` to Gradle Project

Stefterv:
The LSP is currently part of the java mode, in the future we intent to separate the LSP from java mode and the PDE in general and this is a first step that people can develop on top of


Stefterv:
This one can be closed as it has become clear that the LSP is not really something that yet _can_ and _needs_ to stand on its own. Ideally in the future we can make a standalone version that can be published to maven but right now it is not important

PR #963, opened 2025-03-07
New Welcome Screen

Stefterv:
Adds a new welcome screen to the PDE build on the new Jetpack Compose system. 

Screenshot 2025-03-07 at 16 50 39


Stefterv:
Converting to draft as the Preferences implementation either needs to be refactored or tested

PR #968, opened 2025-03-11
Gradle Distribution Setup

Stefterv:
I have modified the GitHub workflows so that we can distribute Processing through gradle.

This should be currently on par with the old ant distribution or better.

- Improved .dmg customisation
- Added Signing and Notarisation on macOS
- Improved .msi customisation
- Changed linux distribution from .deb to .snap for more widespread support
- Added automatic deployment to the Snap Store
- General quality of life and code improvements when it comes to the GitHub Actions
- Added a compiled version of the website as the downloadable for reference (in the future we should make Processing download the appropriate reference for each version)
- Shimmed the `getenv` and `setenv` methods in `Platform` to not modify environment variables at runtime
- [x] Setup `PROCESSING_SIGNING` & `PROCESSING_SIGNING_IDENTITY` secrets
- [x] Setup `SNAP_NAME` & `GRADLE_GROUP` variables
- [x] Add Windows installer customisation
- [x] Add macOS installer customisation
- [x] Setup Snap Store 
- [x] Disable the `ant` based release workflow for now


SableRaf (approved):
Looks good to me! Thanks Stef

PR #982, opened 2025-03-17
Fixed a situation where larger base64 sketches didn't decode properly

Stefterv:


SableRaf:
> ?.map { it.split("=", limit = 2) }

Only split `pde:` URIs on the first `=` and not on ones that are part of the Base64 encoded data

SableRaf (approved):

PR #1001, opened 2025-03-24
Add Smooth Scrolling to the PDE

Stefterv:
In an effort to make the PDE more modern feeling, I realised this morning that a big part of the issue is that the scrolling updates per line, not with the scroll wheel (mostly with smooth scrolling on macOS).

This change will make scrolling in the PDE fractional, meaning you can scroll 1/3 or a line or less if you want. It also increases the re-rendering frequency of scrolling updates.

## Demo 

https://github.com/user-attachments/assets/aa895496-95a6-4eab-9168-1c576c5be18e

## Left to do

- [ ] Fix an issue where the gutter is not drawn if you scroll fast enough
- [ ] Fix an issue where some of the redrawing is not correct
- [ ] Fix an issue where the user can overscroll the sketch (although this could also be a feature)
- [ ] Horizontal smooth scrolling
- [ ] Fix Scrolling by scrollbar
- [ ] Implement dampening on Windows and Linux?

### DIFF
--- a/app/src/processing/app/syntax/JEditTextArea.java
+++ b/app/src/processing/app/syntax/JEditTextArea.java
@@ -172,33 +172,23 @@ public class JEditTextArea extends JComponent
 //    focusedComponent = this;
 
     addMouseWheelListener(e -> {
-      if (scrollBarsInitialized) {
-        if (e.getScrollType() == MouseWheelEvent.WHEEL_UNIT_SCROLL) {
-          int scrollAmount = e.getUnitsToScroll();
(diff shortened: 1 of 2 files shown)


SableRaf:
Hi Stef, just checking to understand the closure. Was this mainly because the PR had gone stale, or were there specific technical concerns with the approach? Asking mainly so we have context if someone wants to revisit this topic in the future.

Stefterv:
It always was a PoC with a lot of ramifications. Looking at it with current knowledge it would break so many things and be hard to get right. A better approach would be to wait for a new `TextArea`

PR #1004, opened 2025-03-24
Hide scrollbars & Version indicator

Stefterv:
- Hiding the scrollbars in the interface unless there is something to scroll
- Moved version indicator from the title to the bottom tab bar

Screenshot 2025-03-24 at 17 03 04


SableRaf:
How will this look when the update indicator is active?

image

Stefterv:
Screenshot 2025-03-24 at 18 40 11

It will look like this

Stefterv:
@SableRaf I added functionality that will copy system info to clipboard if you click the version number, let me know if you think that would be useful

Also clicking it 5 times will enable DEBUG mode (logs etc) until restart

SableRaf:
I love it! 🐴👍

SableRaf (approved):

PR #1006, opened 2025-03-24
Visual fixes for Linux

Stefterv:
Screenshot from 2025-03-24 19-31-25
Applying the same visual tweaks from windows on linux


Stefterv:
Screenshot 2025-03-24 230638
Fixed the issue where if the window loses focus the title becomes harder to read

SableRaf (approved):

PR #1018, opened 2025-03-26
Fix a color issue on macOS

Stefterv:
On a Mac with an HDR display it would render colours incorrectly in the PDE

Screenshot 2025-03-26 at 16 50 14


SableRaf:
Which one is before? 😀

Stefterv:
The one on the left

SableRaf (approved):

PR #1022, opened 2025-03-28
Store where Processing is installed

Stefterv:
It has been clear to me for a while that third-party tools need a way to discover where Processing is installed.

The Java Preferences are not the best solution but they are the best documented and clear on how this would work and is a tool that we won't have to maintain.

- [ ] Add some documentation to the repo on how to read the Java preferences outside of Java
- [ ] Remove old install locations
- [ ] Remove duplicate install locations because version changed


SableRaf (approved):
lgtm, let's open an issue for the remaining todos

PR #1038, opened 2025-04-15
Change the default display density

### LINKED ISSUE #950: Set default pixel density to the display's density

Stefterv:
Fixes #950 

Changes the default sketch density to reflect the current display in use.

### DIFF
--- a/core/src/processing/core/PApplet.java
+++ b/core/src/processing/core/PApplet.java
@@ -10105,6 +10105,8 @@ public class PApplet implements PConstants {
     sketch.present = present;
     sketch.fullScreen = fullScreen;
 
+    sketch.pixelDensity = sketch.displayDensity();
+
     // For 3.0.1, moved this above handleSettings() so that loadImage() can be
     // used inside settings(). Sets a terrible precedent, but the alternative
     // of not being able to size a sketch to an image is driving people loopy.


SableRaf (approved):

core/src/processing/core/PApplet.java
```diff
@@ -10105,6 +10105,8 @@ static public void runSketch(final String[] args,
     sketch.present = present;
     sketch.fullScreen = fullScreen;
 
+    sketch.pixelDensity = sketch.displayDensity();
```
agreyson, line 10108: I stumbled across this just now while helping a new CS teacher come up to speed on one of my former class projects. The project uses a nested loop and the width and height system variables to change the color of each pixel in the pixels array (as described in chapter 15 of Shiffman's "Learning Processing"). As of this change, that same loop only modifies the pixels in a small portion of the display window on my high DPI laptop. It's a nasty little bug if you haven't faithfully kept up with the Processing release notes. Perhaps you could display a warning message to ease folks into it?

SableRaf: Hi @agreyson Thanks for your comment! We made this change because sketches looked blurry on high-DPI displays, but as you found out, it can break older code that works with pixels directly. This happens because different pixelDensity values introduce inconsistencies in per-pixel operations. @catilac and @tychedelia are exploring solutions here: https://github.com/processing/processing4/pull/1145, and suggestions are welcome on the issue thread: https://github.com/processing/processing4/issues/1131. Since the fix is more complex than expected, adding a warning message could be a helpful interim step to prevent the headaches you ran into. Thanks for your suggestion.

agreyson: Thanks @SableRaf. I appreciate the speedy response! Given that this breaks a substantial chunk of @shiffman instructional materials, my recommendation would be to revert #1038 for now, and leave it to knowledgable developers to call pixelDensity explicitly until you've found a solution (easier said than done, judging by the commentary). Thank you for your work on Processing, and best of luck!

PR #1072, opened 2025-05-04
Terminate JVM of already running instances

### LINKED ISSUE #1066: Second app instance does not exit

PDE

### Processing version

4.4.3 (portable)

### Operating system

Linux (Ubuntu 20.04 x64)

### Steps to reproduce this

"1. Download `processing-4.4.3-linux-x64-portable.zip`, extract it, and `cd Processing/bin`

2. `./Processing &` and wait for Processing to launch.

3. `./Processing &` again and wait for the second Processing IDE to launch.

4. Close both Processing windows.

5. `ps | grep Processing` and you'll see the second Processing process is still alive."

### snippet

```processing
```

### Additional context

I also observed the same happens for local build with Gradle.
The problem is not critical since `Ctrl-C` terminates it anyway, but it is a bit bothering when it happens.

No; I am not familier with Kotlin, sorry.

AhmedMagedC:
## Changes 
added `System.exit(0)` 
**why?**
when new Processing instance opens, a client socket is instantiated and connected to the main server socket (initiated by the main instance), then the main instance handles opening the sketch
but what about the instance that created the socket in the first place? we didn't terminate it's JVM, so it remains consuming resources and doing pretty much nothing

## Tests
i validate the changes by trying to locally build multiple instances
everything works as expected, and these instances automatically exits

### DIFF
--- a/app/src/processing/app/Base.java
+++ b/app/src/processing/app/Base.java
@@ -324,6 +324,8 @@ public class Base {
 
 //      long t10 = System.currentTimeMillis();
 //      System.out.println("startup took " + (t2-t1) + " " + (t3-t2) + " " + (t4-t3) + " " + (t5-t4) + " " + (t6-t5) + " " + (t10-t6) + " ms");
+    } else {
+      System.exit(0);
     }
   }
 
--- a/app/src/processing/app/Base.java
+++ b/app/src/processing/app/Base.java
@@ -251,7 +251,16 @@ public class Base {
 
 //    long t2 = System.currentTimeMillis();
 
-    if (DEBUG || !SingleInstance.alreadyRunning(args)) {
+    // boolean flag indicating whether to create new server instance or not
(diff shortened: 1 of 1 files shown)


Stefterv:
Hi @AhmedMagedC Looks good! 

Code wise, would you be up for changing the code-flow that it becomes clearer when System.exit(0) is called?
Right now I'm a bit worried that it'll be missed the next time changes are made since the check is on line 254 and the action is on line 327

Stefterv (approved):

AhmedMagedC:
Done! what do you think [@Stefterv](https://github.com/Stefterv)?

Stefterv:
Perfect! thank you very much

SableRaf:
Thanks for your contribution @AhmedMagedC 💙

@all-contributors please add @AhmedMagedC for code

PR #1082, opened 2025-05-07
incorrect uniform resolution

### LINKED ISSUE #1063: Some `shader()` examples behave unexpectedly at higher pixel densities

OpenGL

### Processing version

4.4.3

### Operating system

Tested on macOS 13.4

### Steps to reproduce this

1. Open Processing 4.4.3

2. Run examples in `Topics/Shaders` (for example Conway);

3. Notice the rendering is wrong

### snippet

n/a

### Additional context

Since version 4.4.3, the default `pixelDensity` is set to match the display's density. During testing, we found that some shader examples break on high-density displays because they implicitly relied on `pixelDensity` being set to 1.

Some investigation in how `shader()` is implemented and what assumptions are made there might be needed.

This can be assigned

AhmedMagedC:
## Changes
correctly set the uniform resolution to reflect the changes in pixelDensity

**Why?**

when pixelDensity is set to 2 either implicitly (starting from Processing 4.4.3 version) ,or when called when creating a sketch
the physical resolution (width and height) is doubled, but that isn't reflected on the `uniform resolution` (the rendering resolution)
so for example in `conway.glsl` this piece of code `vec2 position = ( gl_FragCoord.xy / resolution.xy );` results in inconsistent position (wrong scaling) because `uniform resolution` still uses the logical resolution specificed in `size()` function

**Tests**
i tested all shaders examples, and everything seems fine except for two examples
1. **deform**, mouseX and mouseY also isn't reflected the change of pixelDensity
a simple solution is to multiple mouseX and mouseY by pixelDensity to reflect the correct position

2. **DemoProjection** this one is missed up, but i think it's related to the code written in sketch not sure about that

### DIFF
--- a/core/src/processing/opengl/PShader.java
+++ b/core/src/processing/opengl/PShader.java
@@ -1210,8 +1210,8 @@ public class PShader implements PConstants {
     }
 
     if (-1 < resolutionLoc) {
-      float w = currentPG.viewport.get(2);
-      float h = currentPG.viewport.get(3);
+      float w = currentPG.viewport.get(2) * currentPG.pixelDensity;
+      float h = currentPG.viewport.get(3) * currentPG.pixelDensity;
       setUniformValue(resolutionLoc, w, h);
     }
 


AhmedMagedC:
Hi [@Stefterv](https://github.com/Stefterv), how are you?
if you could please provide your feedback on this PR
for the **DemoProjection**  example, it uses other uniforms like `transform` and `modelView`, so maybe more investigation is required of how  these are being set and if they adapt the change of pixel density or not
so what do you think?

tychedelia:
I'm not 100% sure about the other implications here, but viewports are typically expressed in terms of physical pixels. We may want to change `PGraphicsOpenGL.setViewport` to account for `pixelDensity` instead of only when writing the resolution uniform buffer. It loos like the only other uniform that uses this is the viewport uniform itself.

catilac:
Closing due to inactivity

Issue #1041, opened 2025-04-16
Improvements and Next Steps for the Processing Language Server (LSP)


SableRaf:
We’re revisiting the Processing Language Server with the goal of improving support for writing and editing Processing code outside the PDE.

### Background

Several contributors have laid important groundwork for this effort:

-   Initial work by @kgtkr was [incorporated for Processing 4.1](https://github.com/processing/processing4/issues/564). They also built a minimal [Processing LSP client for VSCode](https://github.com/kgtkr/processing-language-server-vscode).
-   Further work by @efratror was included in Processing 4.3 (see https://github.com/processing/processing4/issues/678 and https://github.com/processing/processing4/issues/690)
-   As part of the `pr05` grant in 2024, @diyaayay worked on another [VS Code extension](https://github.com/diyaayay/processing-language-server-extension) though it did not make use of the built-in LSP.

### Proposed Improvements

We’d like to improve and extend the current functionality of the built-in LSP.  Priorities are:

- [ ]    Display parameter names in function signatures  (e.g., `copy(sx, sy, sw, sh, dx, dy, dw, dh)`) should show named parameters
- [ ]    "Go to declaration" for built-in functions and classes  (linking to Processing source code when possible)
- [ ]    Hover documentation for built-in functions and core classes
- [ ]    Support for overloaded functions  (e.g., `colorMode(mode)`, `colorMode(mode, max)`, `colorMode(mode, max1, max2, max3)`)

If you're interested in contributing or discussing further, feel free to comment here. 

Thanks again to everyone who has contributed to this work so far!

Gmin2:
Hey @SableRaf would love to work on Displaying parameters name in the function signatures functionality, i can go ahead with it right?

SableRaf:
Hi @Gmin2! Nice to hear from you 😃 Sure, if you'd like to make a PR, we'll be happy to review it.

Gmin2:
the build is broken for the lsp server when i create a client and connect to the lsp server, the server does not compile, should i show the error logs here ?

SableRaf:
Hi @Gmin2! Thanks for looking into it. Please do share the logs, thanks.

A quick note: with 4.4.0 and 4.4.1 the internal layout of the Processing app changed somewhat, breaking the `processing-java` CLI. The issue you're having might be related. We'll be releasing a fix in 4.4.3 tomorrow. Please use this PR instead in the meantime https://github.com/processing/processing4/pull/1046

cc @Stefterv

[...]

PR #1086, opened 2025-05-09
feat: added parameters named in parameters function

Gmin2:
Added parameters  names in function signatures

Part of #1041 

Some examples
```
{
      "label": "fill",
      "kind": 3,
      "detail": "fill(r, g, b, a)",
      "insertText": "fill($1)",
      "insertTextFormat": 2
}
```

```
{
      "label": "background",
      "kind": 3,
      "detail": "background(x, y, width, height)",
      "insertText": "background($1)",
      "insertTextFormat": 2
}
```

```
    {
      "label": "copy",
      "kind": 3,
      "detail": "copy(sx, sy, sw, sh, dx, dy, dw, dh)",
      "insertText": "copy($1)",
      "insertTextFormat": 2
    }
```

### DIFF
--- a/java/src/processing/mode/java/lsp/PdeAdapter.java
+++ b/java/src/processing/mode/java/lsp/PdeAdapter.java
@@ -5,6 +5,7 @@ import java.net.URI;
 import java.util.AbstractMap;
 import java.util.Arrays;
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ExecutionException;
 import java.util.HashSet;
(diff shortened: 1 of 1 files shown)


java/src/processing/mode/java/lsp/PdeAdapter.java
```diff
@@ -440,4 +473,101 @@ static class Offset {
       this.col = col;
     }
   }
+
+  // Map of known Processing methods and their parameter names
+  private static final Map<String, String[]> methodKnownParameterNames = new HashMap<>();
+  static {
+    // PApplet drawing methods
+    methodKnownParameterNames.put("rect", new String[]{"x", "y", "width", "height"});
```
Stefterv, line 481: This information should be gathered dynamically, not with a pre-defined list

Stefterv (changes requested):
Hi @Gmin2 Thank you for your efforts on this! 

We will need something that will figure out the parameters from core instead of a pre-defined list. In this case we would have to maintain the parameter definitions in the future and those might get out of sync over time.

PR #1091, opened 2025-05-12
Fix setResizable(true) bug on Linux Mint

### LINKED ISSUE #1003: Resizable Window has rendering Bug

Core/Environment/Rendering

### Processing version

1295

### Operating system

Linux Mint 22.1 x86_64; Kernel: 6.8.0-54-generic

### Steps to reproduce this

1. Download Source Code from https://github.com/processing/processing4/releases/tag/processing-1297-4.3.4 

2. Open Project in IntelliJ and run Ant Target Linux (See Picture in Additional context for Configuration)

3. Paste Code from the Example of setResizable() into the IDE and run (See Picture in Additional context)

4. Resize the window

### snippet

```processing
void setup() {
  size(200, 200);
  surface.setTitle("Hello World!");
  surface.setResizable(true);
  surface.setLocation(100, 100);
}

void draw() {
  background(204);
  line(0, 0, width, height);
  line(width, 0, 0, height); 
}
```

### Additional context

Image

Image

No

duskvirkus:
Fixes #1003 

Tested this after a gradle clean. Ran it about 8-10 times to ensure it was working as expected.

Screenshot from 2025-05-12 12-44-51

I hope the reasoning behind the change is explained in comments but it seems like for whatever reason sometimes the window manager doesn't like multiple calls to `frame.setResizable()`. Noticed in debugging that if I removed where we set it to `false` then it works as expected. So the solution ultimately was to dispose of the frame and recreate it. Also noticed that a previous section of code about Insets likely would be good to un-comment because I was noticing black bars around sketches occasionally and this block of code seems to solve that.

This pr is intended to be squash merged.

### DIFF
--- a/core/examples/src/main/java/Issue1003.java
+++ b/core/examples/src/main/java/Issue1003.java
@@ -16,12 +16,11 @@ public class Issue1003 extends PApplet {
     }
 
     public void draw(){
-//        ((PSurfaceAWT) surface).setResizable(true);
-
         background(frameCount % 255);
         line(0, 0, width, height);
         line(width, 0, 0, height);
     }
+
     public static void main(String[] passedArgs) {
         String[] appletArgs = new String[]{ Issue1003.class.getName()};
         if (passedArgs != null) {
--- a/core/examples/src/main/java/Issue1003.java
+++ b/core/examples/src/main/java/Issue1003.java
@@ -1,0 +1,34 @@
+import processing.core.PApplet;
+import processing.awt.PSurfaceAWT;
+
+// Reproduction of issue #1003
+// Resizeable causes crash on linux mint
+public class Issue1003 extends PApplet {
+
+    public void settings(){
+        size(200, 200);
+    }
+
+    public void setup() {
+        ((PSurfaceAWT) surface).setTitle("Hello world!");
+        ((PSurfaceAWT) surface).setResizable(true);
+//        surface.setLocation(100, 100);
+    }
+
+    public void draw(){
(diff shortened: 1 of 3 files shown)


SableRaf:
Hi @duskvirkus and thanks for your efforts on this PR.

I did notice a couple issues in testing (some of those may not be new). Note: I only tested on macOS so feel free to ignore those in case they are not relevant to the issue on Linux Mint. I can also open a new issue.

### 1. Glitchy rendering when resizing below the minimum size

A notable issue: when dragging inside the window beyond the minimal size, rendering starts glitching out. 

Note: this issue is already present in 4.4.3

https://github.com/user-attachments/assets/49dcf904-a89c-42a9-91d9-c0e867770f8c

---

### 2. Low resolution after exiting fullscreen

If you set the window to fullscreen (using the green maximize button on macOS), then press the green button again to return the sketch to its windowed state, the sketch gets very pixelated.

Note: this issue does not exist in 4.4.3 so it seems to be a regression from this PR.

Edit: I'm now unable to reproduce this issue.

image

duskvirkus:
Thanks for the feedback. I took a quick look at some of these issues and was able to reproduce the first issue.

I could not find a quick fix for it so it I'll have to continue working on it over the coming days or weeks when I've got some free time.

I have yet to reproduce the second one but I will continue trying. It seems like it might be a infrequent bug so not sure if it was produced in this pr if it only happens sometimes. Unfortunately I don't have a mac to see if it is platform specific.

catilac:
Closing due to inactivity.

Issue #1099, opened 2025-05-15
Move `serial` Library to Gradle


Stefterv:
### Most relevant area for this enhancement?

Build process

### Feature to enhance

Build system

### Desired enhancement

In PR #888 I took a little shortcut with the existing Java based libraries and made them still reliant on the old Ant build system. 

We would need the ant based `build.xml` to be replaced by a proper `build.gradle.kts` version to further remove our reliance on Ant

### Would you like to work on the issue?

This is a great place to start contributing for someone else

https://github.com/processing/processing4/tree/main/java/libraries/serial

madhav2348:
Hey @Stefterv, can I work on this issue? As a Beginner, this is a good starting point for me

PR #1103, opened 2025-05-16
Migrate ant based `serial` library to Gradle

madhav2348:
### Description
Migrated the serial library from Ant-based `build.xml` to Gradle `build.gradle.kts`

 ### Resolve Issue

#1099 

### Changes 
 Migrate Ant-based `build.xml` with `build.gradle.kts` for java/libraries/serial

### Checklist
 Build succeeds using Gradle
 serial no longer relies on Ant

- `build.gradle.kts` creates `build` folder in `library/serial` by default contains `serial.jar`, even if define source outDir path explicitly
- If I deleted `build.xml`, the Branch Builds (Legacy) GitHub action failed as `build.xml` was not found

Although I was able to run the project and only commit the main Gradle file changes

Let me know if further adjustments are needed!

### DIFF
--- a/java/libraries/serial/build.gradle.kts
+++ b/java/libraries/serial/build.gradle.kts
@@ -27,7 +27,7 @@ tasks.register("checkCore") {
 }
 
 tasks.register<Jar>("buildSerial") {
-    dependsOn("compileSerial")
+    dependsOn("checkCore")
     archiveFileName.set("serial.jar")
     destinationDirectory.set(file("library"))
     from(sourceSets.main.get().output)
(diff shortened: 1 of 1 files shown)


Stefterv:
Hi @madhav2348 Thank you for your contribution and your effort!

Rather than directly translating the Ant build file to Gradle in a one-to-one translation we are looking to have something more Gradle native. Please take a look at how the other gradle files have been organised, especially with regard to importing the results of other modules.

madhav2348:
Thankyou for your feedback , I'll make sure to implement those approach

PR #1106, opened 2025-05-16
Migration of the `serial` library to Gradle

madhav2348:
### Description
Migrated the serial library from Ant-based `build.xml` to Gradle `build.gradle.kts`

 ### Resolve Issue

- #1099 

### Changes 
 -  Added Gradle build logic using Kotlin DSL
 - Removed ant.importBuild("build.xml")
 - Add `settings.gradle.kts`

Let me know if further adjustments are needed!


Stefterv:
Hi @madhav2348 I have opened #1189 as an example of how this kind of migration should be approached. Would you be interested in adjusting your approach based on that PR?

madhav2348:
Sure , I'll be happy to work on that

madhav2348:
hi @Stefterv , taking your example as reference, I searched `jssc` in Maven repository for the serial library :
```js
dependencies {
    compileOnly(project(":core"))
    implementation("org.scream3r:jssc:2.8.0")  
}
```
But it doesn't have the method `getPortProperties()`used in :
https://github.com/processing/processing4/blob/3f36db5742357023270ae3ad2daa97c68c79531e/java/libraries/serial/src/processing/serial/Serial.java#L262-L265

Somehow I found this issue https://github.com/processing/processing/pull/2093/commits/f0cc8759b804e51f1c707e021fcd22d410e5f2a5

now [gohai/java-simple-serial-connector](https://github.com/gohai/java-simple-serial-connector) is only a repo, but doesn't have any release or  `.jar` file that I could use [Jitpack](https://jitpack.io)
and also there is no previous release I could find in the original repo


the solution I thought is
 - use the `jssc.jar` in the library directory
 
or fork the repo and build/upload the .jar file in release 

But I need your advice on this: if I'm doing it right!

Thankyou

Stefterv:
Hi @madhav2348 thank you for flagging this and doing some of the initial research!

Looking [at the history](https://github.com/processing/processing4/commit/fce6a5bfe8947e1ec63186439baf4ad87005149b) it seems that the last change reverted it back to a version of https://github.com/java-native/jssc rather than a custom build. Please check if the latest (2.10.2) works or try to find out which version is currently in the repo. 

Otherwise just use an `implementation()` tag pointing at the existing `.jar` and add a TODO pointing to https://github.com/java-native/jssc and create a new issue here in the repo to request migration of jssc to a supported build on maven.

madhav2348:
would creating an issue be the correct step ? because it is customized by a previous contributor and its depending for many years 
latest version (2.10.2)  and other previous version not working of https://github.com/java-native/jssc , I searched most of releases
https://github.com/processing/processing4/blob/3f36db5742357023270ae3ad2daa97c68c79531e/java/libraries/serial/library/jssc.txt#L1

Stefterv:
We should create an issue on this repo to start work on creating a more maintainable solution for this jar, either that will be incorporating the project into the repo or creating a downstream repo on the foundation org with the changes. We can figure that out in the new issue

Stefterv (approved):

java/libraries/serial/build.gradle.kts
```diff
@@ -1 +1,36 @@
...
+    }
+}
+
+repositories {
+    mavenCentral()
+}
+
+dependencies {
+    compileOnly(project(":core"))
+    // TODO: https://github.com/java-native/jssc
```
catilac, line 19: is this TODO intentional? Or can it be removed?

catilac (approved):
Thank you @madhav2348 !

java/libraries/serial/build.gradle.kts
madhav2348: It is intentional because the serial library depends on JSSC, which has been customized by previous contributors. Unfortunately no publicly version available or maintained that support the current setup

PR #1107, opened 2025-05-17
Add error check for pixelDensity in mask()

### LINKED ISSUE #1065: PImage.mask() broken

Image, Core/Environment/Rendering

### Processing version

4.4.3

### Operating system

macOS

### Steps to reproduce this

Run the following snippet.

Using `mask()` to mask an image leads to `mask() can only be used with an image that's the same size.` though both images are created with same size. Worked until 4.4.2.

### snippet

```processing
PImage img;

void setup() {
  size(150, 150);
  setupHexagon(50);
  hexagon(75, 75, 50);
}

void draw() {
}

private void setupHexagon(int radius) {
  noiseSeed(1004);
  img = createImage(radius*2, radius*2, RGB);
  for (int px = 0; px < img.width; px++) {
    for (int py = 0; py < img.height; py++) {
      float n = noise(px/(float)img.width, py/(float)img.height);
      img.pixels[px+py*img.width] = color(n*255);
    }
  }
  img.updatePixels();

  PShape s = createShape();
  s.beginShape();
  s.fill(255);
  for (int i=0; i<6; i++)
    s.vertex(cos(PI/3*i) * radius, sin(PI/3*i) * radius);
  s.endShape(CLOSE);
  PGraphics maskImage;
  maskImage = createGraphics(img.width, img.height);
  maskImage.beginDraw();
  maskImage.shape(s, img.width/2., img.height/2.);
  maskImage.endDraw();
  img.mask(maskImage);
}

private void hexagon(float x, float y, float radius) {
  imageMode(CENTER);
  image(img, x, y, 2*radius, 2*radius);
}
```

### Additional context

Snippet draws a hexagon with noisy pattern.

assign to someone else, please

pnngocdoan:
Resolves: [#1065](https://github.com/processing/processing4/issues/1065)

Changes:
Add an error check for the mask() function in PImage.java to ensure both pixelWidth and pixelHeight of both imgs are matched.

Tests:
When pixel sizes don't match due to density difference, it will show an error line like below
image

When density doesn't match but underlying pixel size matches:
image

### DIFF
--- a/core/src/processing/core/PImage.java
+++ b/core/src/processing/core/PImage.java
@@ -840,7 +840,6 @@ public class PImage implements PConstants, Cloneable {
    */
   public void mask(PImage img) {
     img.loadPixels();
-    this.loadPixels();
     if (this.pixelWidth != img.pixelWidth || this.pixelHeight != img.pixelHeight) {
       if (this.pixelDensity != img.pixelDensity) {
         throw new IllegalArgumentException("mask() requires the mask image to have the same pixel size after adjusting for pixelDensity.");
--- a/core/src/processing/core/PImage.java
+++ b/core/src/processing/core/PImage.java
@@ -840,6 +840,15 @@ public class PImage implements PConstants, Cloneable {
    */
   public void mask(PImage img) {
     img.loadPixels();
+    this.loadPixels();
+    if (this.pixelWidth != img.pixelWidth || this.pixelHeight != img.pixelHeight) {
+      if (this.pixelDensity != img.pixelDensity) {
+        throw new IllegalArgumentException("mask() requires the mask image to have the same pixel size after adjusting for pixelDensity.");
+      }
+      else if (this.width != img.width || this.height != img.height) {
+        throw new IllegalArgumentException("mask() requires the mask image to have the same width and height.");
+      }
+    }
     mask(img.pixels);
   }
 


core/src/processing/core/PImage.java
```diff
@@ -840,6 +840,15 @@ public void mask(int[] maskArray) {  // ignore
    */
   public void mask(PImage img) {
     img.loadPixels();
+    this.loadPixels();
```
Stefterv: Great work!

Could you please explain to me why the pixels of the current image need to be loaded in before checking the width and height?

pnngocdoan: Hi @Stefterv, so sorry for taking so long to reply, I was out of town last week. I saw the masked img used loadPixels() so I thought it's needed for the current one to loadPixels() to load other pixel properties like width, height, and density. I just checked the function and realized that it just kinda resets the pixels. Thank you so much for pointing that out! Would you like me to remove that line?

Stefterv: Hi @pnngocdoan No worries, appreciate your work! Catching stuff like this is what reviews are for 💙
Yes please remove this line then, thanks for checking thoroughly

pnngocdoan: Just removed it! Thanks for your review!

Stefterv (approved):

SableRaf:
Thanks @pnngocdoan for your contribution! ✨

@all-contributors please add @pnngocdoan for code

SableRaf:
@all-contributors please add @pnngocdoan for code

PR #1118, opened 2025-05-23
Fix library class version error

ThexXTURBOXx:
I depended on a library compiled for Java 21 and stumbled upon this error.
I was quite puzzled by the error message - however after some testing I was able to confirm that Java 17 libraries work fine whilst Java 18 libraries (and later) do not.
This PR fixes the confusing error message.

### DIFF
--- a/java/src/processing/mode/java/runner/Runner.java
+++ b/java/src/processing/mode/java/runner/Runner.java
@@ -775,9 +775,9 @@ public class Runner implements MessageConsumer {
 
     } else if (exceptionClass.equals("java.lang.UnsupportedClassVersionError")) {
       listener.statusError("UnsupportedClassVersionError: A library is using code compiled with an unsupported version of Java.");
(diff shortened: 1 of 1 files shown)


tychedelia (commented):
Might want to see if you can use something like https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/Runtime.Version.html in order to avoid the need to keep this info string up to date.

ThexXTURBOXx:
Done - that's actually a pretty good idea. At first, I thought about `System.getProperty("java.version")`, but `Runtime#version` is way cleaner, I think.

Stefterv (approved):

SableRaf:
Thanks @ThexXTURBOXx and @tychedelia 💙

SableRaf:
@all-contributors please add @ThexXTURBOXx for code

SableRaf:
@all-contributors please add @tychedelia for review

PR #1130, opened 2025-06-01
migrating Preferences and Messages to stand-alone version

### LINKED ISSUE #1104: Making `Preferences` standalone
For the migration of the internal Gradle runner, I'm running into trouble with the `Preferences` class. 

### Problem description

Currently `Preferences` is tightly coupled with the PDE through the use of the `Base`, `Messages`, `Toolkit` and  `Language` classes. These classes do not operate outside of the PDE, thus not allowing reuse of the `Preferences` class in other modules within Processing (e.g. Gradle plugins, Processing CLI, Processing pre-processor, etc.)

### Proposed solution

Migrate the `Preferences` class to a standalone version.

- Completing the migration of Preferences to be standalone from `:app` by moving it to a `:app:utils` module.
- Move functionality dependent on `:app` into a `AppPreferences` class or into other relevant areas of the PDE
- Adding callback events/listeners when a preference changes
- Tests that will declare and test the functionality of `Preferences`, a few of the top of my head would be:
  - Test if the defaults are loaded and saved without creating changes
  - Test if `\` work
  - Test if the preferences file is created if it does not yet exist
  - Any other tests to make sure we have backwards compatibility 
- A Github Action that will run the Tests on all supported platforms
- Bonus: Removing the dependency on `core`

### Steps taken so far
- In IntelliJ IDEA create a new module `utils` in the `app` folder with the following settings (see image below)
- Add package `processing.utils` to the newly created module
- Refactoring the class by right clicking the `Prefrences` class and `Refactor -> Move Class...` and moving it to the newly created `processing.utils` package

Image

AhmedMagedC:
## Changes
- migrating `Preferences` to `:app:utils` module making it independent of `app` module by also moving the functionalities that depends on `app` to `AppPreferences` class
- doing the same for the `Messages` class to be able to use it (for example in debugging) in different areas of Processing
- moving `Utils` into `:app:utils`, because why keeping it in `app` it doesn't need any functionalities from there, and also can be used in `:app:utils` without the need to depend on `app`
- now for the platform-dependent settings location, the previously mention approach in the [issue](https://github.com/processing/processing4/issues/1104) made cyclic dependency problem, so i moved the `getSettingsFolder()` method for each platform from `:app` to `:app:utils`, without duplication the code i made `app.Platform` class to depend on `SettingsResolver` in `:app:utils` to get the correct location for each platform
- loading the `defaults.txt` now by using JAR resource system instead of `Platform.getContentFile()` (which is also marked as deprecated in the codebase encouraging to migrate to JAR RS)

## Todo
- [ ] Adding callback events/listeners when a preference changes
- [ ] Tests that will declare and test the functionality of Preferences
- [ ] A Github Action that will run the Tests on all supported platforms

## Tests
I tried to compile and run, as for now preferences are loaded and saved correctly

but for some reason some languages aren't displayed properly like: arabic , chinese, but languages like: english, japanese properly displayed.
I will look into it, but any help or suggestion of why that happens is very much appreciated

### DIFF
--- a/BUILD.md
+++ b/BUILD.md
@@ -113,7 +113,7 @@ Processing consists of three main components: `Core`, `Java`, and `App`. The `Co
 - If you've written a large sketch and Processing has become slow to compile and run it, a place to improve this code can most likely be found in the `Java` section.
 
 ## User interface
-Historically, Processing's UI has been written in Java Swing and Flatlaf (and some html & css). Since 2025 we have switched to include Jetpack Compose, mostly for it's backwards-compatibility with Swing. This approach allows us to gradually replace Java Swing components with Jetpack Compose ones, instead of doing a complete overhaul of the editor.
+Historically, Processing's UI has been written in Java Swing and Flatlaf (and some html & css). Since 2025 we have switched to include Jetpack Compose. It is backwards-compatible with Swing, which allows us to gradually replace Java Swing components with Jetpack Compose ones, instead of doing a complete overhaul of the editor.
 
 ## Build system
 


AhmedMagedC:
so for the language display issue, i suspect it happened when i renamed `Preferences` class to `AppPreferences`
the comment: `# Preferences (Frame)` in all `PDE.properties` automatically changed to `# AppPreferences (Frame)` leading to corruption of some files (idk why tho)

it's fixed now

Stefterv:
Hi @AhmedMagedC Thank you for the PR! Sorry I got a little busy, hope to get to this soon!

app/src/processing/app/Base.java
```diff
@@ -472,12 +471,12 @@ static public String getVersionName() {
 
 
   public static void setCommandLine() {
-    commandLine = true;
+    System.setProperty("processing.cli", "true");
```
Stefterv: Could you please explain why this needed to change?

app/src/processing/app/Messages.kt
```diff
@@ -26,45 +26,8 @@ import java.io.StringWriter
...
-class Messages {
+class Messages : processing.utils.Messages() {
     companion object {
-        /**
-         * "No cookie for you" type messages. Nothing fatal or all that
-         * much of a bummer, but something to notify the user about.
-         */
-        @JvmStatic
-        fun showMessage(title: String = "Message", message: String) {
-            if (Base.isCommandLine()) {
```
Stefterv: I would say that the `JOptionPane` parts should stay and the println can be in the universal class

app/utils/src/main/java/processing/utils/Preferences.java
```diff
@@ -0,0 +1,308 @@
...
+
+    // next load user preferences file
+    preferencesFile = Base.getSettingsFile(PREFS_FILE);
+    boolean firstRun = !preferencesFile.exists();
+    if (!firstRun) {
+      try {
+        load(new FileInputStream(preferencesFile));
+
+      } catch (Exception ex) {
+        Messages.showError("Error reading preferences",
```
Stefterv: I think these could be transformed to normal errors and then catch and log them in the `app` classes

app/utils/build/classes/java/main/processing/utils/Base.class
Stefterv, line 1: Please add the `build` folder to the .gitignore

Stefterv (commented):
Hi @AhmedMagedC Thank you again for the work! It is heading very much in the right direction. I have added some comments as suggestions to help further improve

app/src/processing/app/Base.java
AhmedMagedC: The main motivation behind this change was to be able to know if Processing is launched via CLI or not without the need to depend on `app` module via using `Base.isCommandLine()`

as for example im using it in the `utils.Messages.kt`

app/src/processing/app/Messages.kt
AhmedMagedC: If i understand you correctly, you wanna seperate the appearance of graphical messages via `JOptionPane` to be only used by `app.Messages` and a simple println only for the `utils.Messages`

but wouldn't that make an issue in `utils` module? as for example in `processing.utils.settingslocation` and `processing.utils.Base`
error messages via `Messages.showError()` wouldn't display a graphical message only just a console print. So is it accepctable? also bear in mind that i cannot use `app.Messages` for this purpose because i want the `utils` module along with the `utils.preference` class to be fully independent of `app` module

app/utils/src/main/java/processing/utils/Preferences.java
AhmedMagedC: I can do that. but that would depend on how we choose to implement `utils.Messages`

if we decided to keep the use of `JOptionPane` in `utils.Messages` then this delegation of catching and logging the errors in `app` rather than in `utils` wouldn't make much sense because we will use the same method but in different location 

if we decided to remove `JOptionPane` and keep only the `println` then i can do that delegation

app/utils/build/classes/java/main/processing/utils/Base.class
AhmedMagedC: already did in this [commit ](https://github.com/processing/processing4/pull/1130/commits/9b9f514aae31dc3b08a198fb8890d7ae435dbdf6)

Stefterv: Yes but the build files themselves are still in the PR (see [here](https://stackoverflow.com/questions/13541615/how-to-remove-files-that-are-listed-in-the-gitignore-but-still-on-the-repositor) for examples on how to clean this)

AhmedMagedC: Okay done!
Thanks for pointing this out, i though git would remove it automaticlly after adding it to .gitignore 
what about the other points above?

Stefterv:
Hi @AhmedMagedC Sorry it took me so long to take a look at your PR again! and again thank you for your efforts, it's helping me a lot trying figure out how to refactor some of these intertwined elements.

Looking at your PR and #1164 I think it is important moving forward that we only migrate one class at a time. I would suggest that any functionality in the `Preferences` class that is currently causing you to have to migrate the `Messages` and `Base` classes for example, should be taken out of the `Preferences` class. That will make the code a little bit more flexible in the future.

AhmedMagedC:
> Hi @AhmedMagedC Sorry it took me so long to take a look at your PR again! and again thank you for your efforts, it's helping me a lot trying figure out how to refactor some of these intertwined elements.
> 
> Looking at your PR and #1164 I think it is important moving forward that we only migrate one class at a time. I would suggest that any functionality in the `Preferences` class that is currently causing you to have to migrate the `Messages` and `Base` classes for example, should be taken out of the `Preferences` class. That will make the code a little bit more flexible in the future.

Hi @Stefterv , It's okay it's also the first time for me trying to refactor a codebase by decoupling tightly-coupled classes

im not sure i follow you,
if i were to move `Messages` back to `app` module, how might we decouple `Preferences` from using it?

also for the change of getting the setting folder for each OS is now independent of `Base` class, i dont think i should undo this change it will make `Preferences` impossible to be a standalone class

Stefterv:
Yes exactly, instead of migrating `Messages` along with `Preferences`, `app.utils.Preferences` should throw an error that is caught within the encapsulating `app.Preferences`, same with the paths `app.utils.Preferences` could take a path as an argument on creation and that argument is set in `app.Preferences`

AhmedMagedC:
Hi @Stefterv , Sorry for the late reply i was busy for the last couple of weeks

i see now that we should currently focus on migrating `Preference` class
i migrated `Messages` with it in the first place to be able to print the error messages
but as you suggested i can simply throw them then caught and handled in `app.preferece` class

for the paths, i dont undetstand why i should undo it
now i fully docoupled getting the preference file location from the app module
you can simply call `SettingsResolver.getSettingsFolder();` which will get you the location without the need to depend on app module

Stefterv:
Hi @AhmedMagedC! Again your work on this has been invaluable in figuring out a strategy for starting to untangle some of these parts. 

I’m recommending we tackle one class at a time to avoid regressions and the risk of having to undo a PR because of an unforeseen side effect in the existing code.

For me, an important part of these decoupling steps is making the coupling between classes more visible. That way, we can gradually refine each class to have clearer inputs and outputs.

PR #1145, opened 2025-06-12
Ensure logical coordinates are used for all operations

### LINKED ISSUE #1131: Regression in 4.4.3: get() and set() are reading and writing from/to the wrong coordinates.

Core/Environment/Rendering

### Processing version

4.4.3

### Operating system

MacOS 11.7.10

### Steps to reproduce this

1. Run the attached script to demonstrate the behavior.

In v4.4.2, the color that get() reads is white, as expected. And set() draws around (350, 350) as expected.  But in v4.4.3 (and v4.4.4) get() reads blue, and set() draws around (175, 175).

It appears that both get() and set() are operating on (x/2. y/2) instead of (x,y).

### snippet

```processing

////////////////////////////////////////////////////////////////////////////////////////////////////////////// //<>// //<>// //<>// //<>//
// Define colors
color white = color(255);
color black = color(0);
color red = color(255, 0, 0);
color blue = color(0, 0, 255);

//////////////////////////////////////////////////////////////////////////////////////////////////////////////
// Setup
void setup() {
    size(400, 400);
    background(white);
    noLoop();
}

//////////////////////////////////////////////////////////////////////////////////////////////////////////////
// Draw
void draw() {
    stroke(blue);
    fill(blue);
    rect(165, 165, 20, 20);

    // BUG 1: The color c is blue but it should be white.
    // It seems to be reading from (x/2, y/2) = (175, 175) instead of (350, 350).
    color c = get(350, 350);
    showColor(c);            // R/G/B = 0.0 0.0 255.0 = blue

    // BUG 2: This draws around (175, 175). It should draw around (350, 350).
    for (int x = 340; x < 360; x++) {
        for (int y = 340; y < 360; y++) {
            set(x, y, red);
        }
    }

    //save("screenshot.xxx.png");
}

//////////////////////////////////////////////////////////////////////////////////////////////////////////////
// Show information about a color
void showColor(color c) {
    print("R/G/B =", red(c), green(c), blue(c) );
    if (c == white) {
        println(" = white");
    } else if (c == blue) {
        println(" = blue");
    } else {
        println("");
    }
}
//////////////////////////////////////////////////////////////////////////////////////////////////////////////

```

### Additional context

_No response_

No

tychedelia:
# Problem

Fixes #1131 

When `pixelDensity`, `get`/`set` operations no longer work on logical but instead physical pixels. This is confusing for users who think that their image is `size` pixels. This behavior will become particular problematic when fractional scaling is introduced as it's more difficult to reason about that size * 2.

# Solution

Ensure that operations are always done in logical pixels, upscaling/downscaling into physical pixels when doing write operations to the backing texture.

Add a new pixel access mode constant that determines whether linear or nearest neighbor sampling should be done when scaling. This isn't super helpful right now but will become more important when fractional scaling is introduced.

To be honest, I'm still not 100% sure this is the right solution. It's clear that the `PImage` api was always intended to be physical pixels and indeed we now introduce a weird distinction where base `PImage` will continue to use physical pixels since it doesn't make sense to refer to images as having a scaling factor. It's probably good for users to not have to know about these concepts but also introduces inherent confusion when using high dpi scaling. More robust documentation could probably help here but I'm still a little apprehensive as this is a substantial change (i.e. a totally new contract that we always use logical pixels).

# TODO:
- Check whether off screen rendering still works
- Confirm whether we need to do gamma correction when doing software rendering in `PGraphicsJava2D`
- Test more examples to see if anything broke
- Double check if there's a better existing pattern for the gpu copies when binding the output texture is necessary
- Double check depth writes still work + msaa can be configured

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -115,3 +115,4 @@ java/build/
 /app/windows/obj
 /java/gradle/build
 /java/gradle/example/.processing
+.kotlin/
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -353,7 +353,14 @@ tasks.register<Copy>("includeJavaMode") {
     duplicatesStrategy = DuplicatesStrategy.EXCLUDE
 }
 tasks.register<Copy>("includeJdk") {
-    from(Jvm.current().javaHome.absolutePath)
+    from(Jvm.current().javaHome.absolutePath) {
+        // TODO: Check if this is still needed when upgrading from JDK 17
+        // https://github.com/adoptium/adoptium-support/issues/937
+        if (OperatingSystem.current().isMacOsX) {
+            exclude("**/*.jsa")
+            exclude("**/legal/**")
+        }
+    }
     destinationDir = composeResources("jdk").get().asFile
 }
 tasks.register<Copy>("includeSharedAssets"){
--- a/core/src/processing/awt/PGraphicsJava2D.java
+++ b/core/src/processing/awt/PGraphicsJava2D.java
@@ -2777,17 +2777,52 @@ public class PGraphicsJava2D extends PGraphics {
 
   @Override
   public void loadPixels() {
-    if (pixels == null || (pixels.length != pixelWidth*pixelHeight)) {
-      pixels = new int[pixelWidth * pixelHeight];
-    }
+    if (pixelDensity > 1) {
+      loadLogicalPixels();
+    } else {
+      if (pixels == null || (pixels.length != pixelWidth*pixelHeight)) {
+        pixels = new int[pixelWidth * pixelHeight];
+      }
 
+      WritableRaster raster = getRaster();
+      raster.getDataElements(0, 0, pixelWidth, pixelHeight, pixels);
+      if (raster.getNumBands() == 3) {
+        // Java won't set the high bits when RGB, returns 0 for alpha
+        // https://github.com/processing/processing/issues/2030
+        for (int i = 0; i < pixels.length; i++) {
+          pixels[i] = 0xff000000 | pixels[i];
+        }
+      }
+    }
+  }
+  
+  protected void loadLogicalPixels() {
+    int logicalPixelCount = width * height;
+    int physicalPixelCount = pixelWidth * pixelHeight;
+    
+    if (pixels == null || (pixels.length != logicalPixelCount)) {
+      pixels = new int[logicalPixelCount];
+    }
+    
+    // Create temporary physical pixel buffe
+    int[] physicalPixels = new int[physicalPixelCount];
+    
     WritableRaster raster = getRaster();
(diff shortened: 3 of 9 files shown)


SableRaf:
Hi @tychedelia and thanks for your work on this so far! 

It seems to properly address the issue for `get()` and `set()` as described in #1131, however I also tried it on some of the examples that broke with the `pixelDensity` change. I hope this is helpful.

In particular, the following shader examples are still broken with this PR:

- `Topics > Shaders > Conway`
- `Topics > Shaders > DomeProjection`
- `Topics > Shaders > Landscape`

There might be other broken examples but these are the ones I remember trying before.

I also tried the example `ControlP5colorWheel` from the ControlP5 contributed library, which now crashes with the following error (that's new with this PR):

```
updatePixels(x, y, w, h) is not available with this renderer.
java.lang.ArrayIndexOutOfBoundsException: arraycopy: last source index 40400 out of bounds for int[40000]
	at java.base/java.lang.System.arraycopy(Native Method)
	at java.desktop/sun.awt.image.IntegerInterleavedRaster.setDataElements(IntegerInterleavedRaster.java:426)
	at processing.awt.PGraphicsJava2D$ImageCache.update(PGraphicsJava2D.java:1791)
	at processing.awt.PGraphicsJava2D.imageImpl(PGraphicsJava2D.java:1600)
	at processing.core.PGraphics.image(PGraphics.java:3911)
	at controlP5.ColorWheel$ColorWheelView.display(Unknown Source)
	at controlP5.ColorWheel$ColorWheelView.display(Unknown Source)
	at controlP5.Controller.draw(Unknown Source)
	at controlP5.ControllerGroup.drawControllers(Unknown Source)
	at controlP5.ControllerGroup.draw(Unknown Source)
	at controlP5.ControlWindow.draw(Unknown Source)
	at controlP5.ControlWindow.draw(Unknown Source)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:77)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:568)
	at processing.core.PApplet$RegisteredMethods.handle(PApplet.java:1342)
	at processing.core.PApplet.handleMethods(PApplet.java:1489)
	at processing.core.PApplet.handleDraw(PApplet.java:2139)
	at processing.awt.PSurfaceAWT$9.callDraw(PSurfaceAWT.java:1388)
	at processing.core.PSurfaceNone$AnimationThread.run(PSurfaceNone.java:356)
```

PR #1156, opened 2025-06-20
Fix JDK File permissions

Stefterv:
Fix issue where including the JDK would not be copyable on subsequent runs of the build processes.


SableRaf (approved):
Looks good to me!

PR #1160, opened 2025-06-26
Update BUILD.md with directions for CLI build

jd-boyd:
Add directions for building Processing from the command line (no IDE) on Ubuntu 24.04.

### DIFF
--- a/BUILD.md
+++ b/BUILD.md
@@ -62,14 +62,32 @@ If your main concern is with the `Core` you don't need to build and start the wh
 
 ## Other editors
 
-### No editor, Ubuntu 24.04
+### No editor, Linux or macOS
+
+The first step is to install Temurin JDB 17 - LTS.
+For Linux on x86, download:
+https://github.com/adoptium/temurin17-binaries/releases/download/jdk-17.0.15%2B6/OpenJDK17U-jdk_x64_linux_hotspot_17.0.15_6.tar.gz
+
+For macOS users on Apple Silicon, download:
+https://github.com/adoptium/temurin17-binaries/releases/download/jdk-17.0.15%2B6/OpenJDK17U-jdk_aarch64_mac_hotspot_17.0.15_6.pkg
+
+For everyone else, visit:
+https://adoptium.net/temurin/releases/?package=jdk&version=17&os=any&arch=any
+
 
 ```
-sudo apt install openjdk-17-jdk openjdk-17-source
-export JAVA_HOME=/usr/lib/jvm/java-17-openjdk-amd64/
+export JAVA_HOME=/path/to/temurin/jdk-17.0.15+6/
 ./gradlew build
-./gradlew tets
+./gradlew test
+```
+
+Then either:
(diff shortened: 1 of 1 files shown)


BUILD.md
```diff
@@ -62,7 +62,15 @@ If your main concern is with the `Core` you don't need to build and start the wh
 
 ## Other editors
 
-TBD
+### No editor, Ubuntu 24.04
+
+```
+sudo apt install openjdk-17-jdk openjdk-17-source
```
Stefterv: We use the Terumin 17 JDK distribution

BUILD.md
```diff
@@ -62,7 +62,15 @@ If your main concern is with the `Core` you don't need to build and start the wh
...
+export JAVA_HOME=/usr/lib/jvm/java-17-openjdk-amd64/
+./gradlew build
+./gradlew tets
```
Stefterv: Typo?

BUILD.md
```diff
@@ -62,7 +62,15 @@ If your main concern is with the `Core` you don't need to build and start the wh
 
 ## Other editors
 
-TBD
+### No editor, Ubuntu 24.04
```
Stefterv: This title can be more general? I think this approach would work more distros than just Ubuntu 24.04 correct?

BUILD.md
```diff
@@ -62,7 +62,15 @@ If your main concern is with the `Core` you don't need to build and start the wh
...
+
+```
+sudo apt install openjdk-17-jdk openjdk-17-source
+export JAVA_HOME=/usr/lib/jvm/java-17-openjdk-amd64/
+./gradlew build
+./gradlew tets
+./gradlew package
```
Stefterv, line 86: I think we should mention that this will create the package in the `app/build/compose/binaries` folder, I think we should also mention Processing can launched directly with `run`

Stefterv (changes requested):
Hi @jd-boyd Thank you for the suggestion! I agree that we should add instructions to build from source. I added some comments on what could be improved

BUILD.md
jd-boyd: Yes.

BUILD.md
jd-boyd: Will update to that then.

BUILD.md
jd-boyd: It can be more general.  As written, it wouldn't work for anything not Debian based, but Temurin seems to prefer being installed by extracting a tarball, so updating the directions to that will make it more general across all linux distros and perhaps macOS.

BUILD.md
jd-boyd: Thank you for those notes, I will add them as well.

jd-boyd:
@Stefterv I believe that I have address your comments.  Thank you for them.

SableRaf (approved):
Looks good to me!

SableRaf:
@all-contributors please add @jd-boyd for documentation

PR #1161, opened 2025-06-28
CLITest for testing CLI commands in Processing IDE

Stefterv:
Easily run Processing command line arguments from IntelliJ IDEA

Screenshot 2025-06-28 at 10 38 29

### DIFF
--- a/app/test/processing/app/CLITest.kt
+++ b/app/test/processing/app/CLITest.kt
@@ -1,0 +1,50 @@
+package processing.app
+
+import java.io.File
+import kotlin.test.Test
+
+/*
+This class is used to test the CLI commands of the Processing IDE.
+It mostly exists to quickly run CLI commands without having to specify run configurations
+or to manually run it on the command line.
+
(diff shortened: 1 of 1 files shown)


Stefterv:
@toniab Is this something that would have helped your work on #1162 ?

toniab:
Hi! This works and is easier than having to edit run config. Is the idea that contributors could replace the arguments there to whatever I want for my testing purposes (and they would make sure not to push their changes to that line?)
-Should we mention this testing functionality in the build or contribute .md's? Do people go to app > test > for other things already?
-Should we add an inline comment explaining that this test is an alternative to editing run config and you can run the test by pressing the corner (to help educate people using it?)

toniab (approved):
Hi! This works and is easier than having to edit run config. 
Is the idea that contributors could replace the arguments there to whatever I want for my testing purposes (and they would make sure not to push their changes to that line?)

Should we mention this testing functionality in the build or contribute .md's? Do people go to app > test > for other things already?

Issue #1138, opened 2025-06-07
processing-java cannot handle paths containing spaces


PhilipSharman:
### Most appropriate sub-area of Processing 4?

Utilities

### Processing version

4.4.4.

### Operating system

MacOS 11.7.10 (Big Sur)

### Steps to reproduce this

1. Install 'processing-java' (via the Tools menu).
2. Create a 'hello world' type sketch.
3. Put it in a folder whose path does _not_ contain spaces.
4. Run e.g. `processing-java --sketch='/Users/philip/Temp/FolderA/hello/' --run`.
5. The sketch runs okay.
6. Copy the sketch and put it in a folder whose path _does_ contain a space.
7. Run e.g. `processing-java --sketch '/Users/philip/Temp/Folder B/hello/' --run`.
8. You get the error `Error: got unexpected extra argument (B/hello/)`.

### Additional context

I tried escaping the spaces in the path (e.g. `processing-java --sketch '/Users/philip/Temp/Folder\ B/hello/' --run`) but I get the same error.

Thanks!

### Would you like to work on the issue?

No

toniab:
Hi Philip, thanks for finding and pointing out this bug. My name is Tonz and I was recommended your issue as a good starting place for contributing to this codebase!

Maintainers:
Can I be assigned this issue?
The bug lies in how the LegacyCLI class was parsing arguments on a per arg basis.
I'd like to submit a fix for this that forwards any submitted arguments as-is to the Commander.

(Once I add that, I will test that this new behavior isn't too broken if no args are attached. If so, are there any args  that -if missing- should trigger warning messages or fallback values etc?)

PR #1162, opened 2025-06-28
Legacy CLI to handle sketch paths with spaces.

toniab:
**Resolves:** [Issue #1138](https://github.com/processing/processing4/issues/1138)

**Changes:** Using arguments.multiple() to pass the argument strings to Commander rather than defining each argument as a Clikt val in advance.

**Tests:** The following configs worked.

- run --args='cli --sketch="/Users/tonz/Documents/Processing/test space/sketch_Test" --run'
- run --args='cli --help'

**Warning:** When testing changes to this file, if you run into an error, try running:
rm -rf /[ YOUR REPOSITORY PATH ]/app/build/compose/tmp 

**Question:** Should we nicely handle if they did not specify a required argument? Right now it errors out if no sketch is passed.


Stefterv (approved):

Stefterv:
Thank you @toniab for this PR! 💙

> Question: Should we nicely handle if they did not specify a required argument? Right now it errors out if no sketch is passed.

Is the error the same as in 4.3.4? If so, then we shouldn't really change it

toniab:
Ok confirmed.
I did createDistributable to compare the PR and main and both give the same message "No sketch path specified." 
(IntelliJ showed deeper error trace so I was confused for a second about what would appear in the terminal).

SableRaf:
Thanks @toniab! ✨

@all-contributors please add @toniab for code

PR #1164, opened 2025-07-03
SketchException refactor

joshgiesbrecht:
Created an `:app:utils` module, moved `SketchException` to a `processing.utils` package within `:app:utils`, changed all references to `SketchException` in both `java` and `app` to use `processing.utils.SketchException`, removed the two duplicate `SketchException` classes in `app` and `java.preproc`.

This currently seems to work (tested only on Linux so far, but I don't think there's anything platform-specific involved), and does solve the bug of highlighting the first error on running a sketch. Non-Java modes will not yet be using the correct SketchException class; I suspect this will be a 'soft' fail that has a similar problem to the existing bug, but I haven't tested this much yet. (Shader mode seems to be broken already?)

Since the draft PR for #1104 also creates `:app:utils` and is more complex than this PR, I'm fine with waiting until the other PR is merged, and then I can make sure this one merges cleanly before it's added in.

What's the right thing to do re: the impact on non-Java modes? Should I submit PRs for those projects as well, maybe once this one's actually merged into main?


joshgiesbrecht:
uhhhh just noticed this PR is re-including commits for the previous bug fix I did. is that a problem, other than being kind of ugly?

joshgiesbrecht:
(Also, I can do some more testing yet, but wanted to check that the merge timing made sense and to ask about the other modes)

((also I like parentheticals too much and am probably overthinking what to say, ha ha))

app/utils/build.gradle.kts
```diff
@@ -0,0 +1,19 @@
+plugins {
+    id("java")
+}
+
+group = "org.example"
```
Stefterv: Please remove these

app/.gitignore
```diff
@@ -1,2 +1,3 @@
 bin
 pde.jar
+/utils/build/
```
Stefterv: Please use the top level .gitignore

Stefterv (commented):
Hi @joshgiesbrecht Sorry it took me so long to take a good look at this. Thank you for taking this on and fixing this issue!

Looks good to me, I added a few comments that are mostly about style.

app/utils/build.gradle.kts
joshgiesbrecht: by "these" you meant the group and version lines, right?

Stefterv: Yep

joshgiesbrecht:
Ok, my git skills are weak, and I can't figure out why this PR is dumping so many excess commits in after I rebased the actual changes to an up-to-date main. The branch on my fork of the repo looks reasonable: https://github.com/processing/processing4/compare/main...joshgiesbrecht:processing4:SketchException-refactor

Should I resubmit the PR? Or do you know of some way to make it smarten up?

(I'm also not sure how this has ended up failing tests, which is not a great sign. It's building here when I checkout the refactor branch.)

Stefterv:
My git knowledge is also not good enough to tell you exactly what you need to do. Feel free to cherry pick your changes onto a new branch and open a new PR. 

As for the failing tests, I checked and it makes sense that they are failing, the `SketchException` class is taken out of a context that Ant (our legacy build system) has access to. One of our next steps is to do away with the Ant based build system. For now I could start disabling the Ant based workflows, what do you think @SableRaf ?

SableRaf:
> For now I could start disabling the Ant based workflows, what do you think @SableRaf ?

That sounds reasonable :)

Issue #1141, opened 2025-06-10
macOS exported app fails with `JRExLoadFullError`


SableRaf:
### Description
A user reported on [Reddit](https://www.reddit.com/r/processing/comments/1l75d2r/cant_open_exported_application_macos_intel/) that exported applications from Processing 4 fail to launch on macOS, showing a `JRExLoadFullError` or `JRELoadError` popup.

_Edit: Exporting on macOS may also throw error `java.util.NoSuchElementException: No value present`. See https://github.com/processing/processing4/issues/1193_

### Cause
The export logic in [`JavaBuild.java`]( https://github.com/processing/processing4/blob/0224d844fe0e78f053e54a92dd051e6131ff813e/java/src/processing/mode/java/JavaBuild.java#L752-L759) looks for a directory that starts with `jdk-`, but in our current setup, the embedded JDK starts with `jdk` (without the dash). This causes the path resolution to fail, and no JRE is bundled, leading to the `JRExLoadFullError`.

### Screenshots
![Image](https://github.com/user-attachments/assets/f04053be-70f3-44a3-bd52-10310c130de1)
![Image](https://github.com/user-attachments/assets/178456d7-6ae4-45a5-b627-c88bc0406f05)

challah:
@SableRaf I'd like to take this on. Is this just about removing the dash in `jdk-` to `jdk`? If so, please assign me this issue. Thank you.

SableRaf:
Thanks Jordanne. Good to see you here! I have assigned the issue to you. Feel free to open a Pull Request and we'll be happy to review it. Let me know if you need any guidance 😃

challah:
It looks like there's an issue beyond the `jdk-`/`jdk` directory path issue for the exports. @Stefterv can you look into this please? Thanks.

Image attached with the `.filter(p -> p.getFileName().toString().startsWith("jdk"))` change and trying to open an exported app. 
![Image](https://github.com/user-attachments/assets/23e04aab-17f8-4d58-a2cb-d1cbf64519fa)

Stefterv:
Just looked into it a bit deeper right now and seems like the launch4j specifically expects an original JDK and not the stripped down version we're shipping to work. This indeed goes a little deeper than I first expected so I'll assign it to myself

JohnRConrad:
Any progress on this?

DW-CN:
I installed the current latest version 4.4.10 but still encountered this issue during Export Applications with Embed Java checked, any comments? Thanks!

[...]

PR #1171, opened 2025-07-09
Fix: Allow JDK folders starting with "jdk" or "jdk-" during export

amankumar100:


SableRaf:
> Summary
> 
> Fixes https://github.com/processing/processing4/issues/1141: macOS exported apps fail with JRExLoadFullError or JRELoadError.
> 
> Cause : The export logic in JavaBuild.java was only checking for directories that start with jdk-, but the embedded JDK can be named just jdk. This caused the bundling step to fail, resulting in missing JREs.
> 
> Fix : Replaced the strict .startsWith("jdk-") check with .startsWith("jdk"), allowing both jdk and jdk- prefixed folders.
> 
> Impact :
> 
> Ensures macOS exported apps bundle the JRE properly
> Prevents runtime errors on macOS after export

Thanks for submitting this patch, and welcome! 

Just a heads-up: we ask that contributors get assigned to an issue before opening a PR. In this case, someone else is already assigned to the issue.

I appreciate the effort you put into this. Since it's a small change, we're happy to review it and share feedback. However, we won't be merging this PR.

Stefterv:
Just tested it on macOS (arm) and removing the dash is not enough to fix the issue
Screenshot 2025-07-09 at 15 11 54

SableRaf:
Thanks for your interest in contributing @amankumar100. Just a reminder: if an issue is already assigned, it means someone is already working on it. You're still very welcome to join the conversation on the issue itself and offer solutions, that kind of input is also valuable. 

If you'd like to contribute code, please start with an unassigned issue. Before you do, make sure to build locally and test your changes. See [BUILD.md](https://github.com/processing/processing4/blob/main/BUILD.md) for instructions.

amankumar100:
> Thanks for your interest in contributing @amankumar100. Just a reminder: if an issue is already assigned, it means someone is already working on it. You're still very welcome to join the conversation on the issue itself and offer solutions, that kind of input is also valuable.
> 

This issue has not been issued to anyone that's why i contributed, also I am contributing through VSCode as per BUILD.md

PR #1189, opened 2025-07-23
Migration of the `pdf` library to Gradle

Stefterv:
Migrates the `pdf` library from Ant to Gradle.

### DIFF
--- a/java/build.gradle.kts
+++ b/java/build.gradle.kts
@@ -1,11 +1,11 @@
 plugins {
-    id("java")
+    java
 }
 
 repositories{
     mavenCentral()
     google()
-    maven { url = uri("https://jogamp.org/deployment/maven") }
+    maven("https://jogamp.org/deployment/maven")
 }
 
 sourceSets{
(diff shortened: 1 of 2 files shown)


java/build.gradle.kts
```diff
@@ -65,8 +67,8 @@ tasks.register<Copy>("copyCore"){
     into(coreProject.layout.projectDirectory.dir("library"))
 }
 
-val libraries = arrayOf("dxf","io","net","pdf","serial","svg")
-libraries.forEach { library ->
+val legacyLibraries = arrayOf("dxf","io","net","serial","svg")
+legacyLibraries.forEach { library ->
```
catilac, line 71: <3

java/build.gradle.kts
```diff
@@ -77,10 +79,30 @@ libraries.forEach { library ->
...
-        into( layout.buildDirectory.dir("resources-bundled/common/modes/java/libraries/$library"))
+        into( javaMode("/libraries/$library"))
+    }
+    bundle.configure {
+        dependsOn("library-$library-extraResources")
+    }
+}
+
+val libraries = arrayOf("pdf")
+libraries.forEach { library ->
```
catilac, line 90: is this because things will move out of the `legacyLibraries` array and into this one?

catilac (commented):
Looks good! Left questions. I'm just going to make sure all of these build as my test? I think that would work, I don't need to actually test using the PDF library? EXCITING TIMES

java/build.gradle.kts
Stefterv: Yes as they are migrated away from Ant

catilac (approved):
this looks good to me. I'm curious about a plugin is defined

PR #1196, opened 2025-08-01
Refactoring `SketchException` to be available outside of `app`

### LINKED ISSUE #1153: No stack trace or thrower position highlighting on exception

PDE

### Processing version

4.4.4

### Operating system

Ubuntu 24.04, Windows 10

### Bug description

The location of an exception (e.g. NullPointerException) is not highlighted, nor is a stack trace printed. The only information printed is a red NullPointerException message at the bottom. There is no obvious way how to find out where this was thrown.

This seems to be a regression to earlier 3.x versions, where the line of the thrower would be at least highlighted. 

### Steps to reproduce this

"1. open the below snippet in PDE

2. run the snippet

3. observe that the error message (NullPointerException) is not linked to any position in the code, nor is any position in the code highlighted, nor is a stack-trace or similar printed anywhere"

### snippet

```
PVector [] vs = new PVector[10];
for (int i=0;i<vs.length;++i)
  vs[i].set(0,0);   // <-- this position should be contained in the error message

```

### Additional context

real showstopper when trying to debug any bigger program in PDE!

No, I’m just reporting the issue

joshgiesbrecht:
Refactored SketchException to live in :app:utils, as per discussion in draft PR https://github.com/processing/processing4/pull/1164. (resubmitted because that draft PR turned into an all-out mess of unrelated commits somehow)

Before merge:

- [x] Disable/remove GitHub actions related to `Ant`

Closes https://github.com/processing/processing4/issues/1153


Stefterv:
Hi @joshgiesbrecht, thank you again for your work on this! I just ran your branch locally and it breaks Android mode since it is looking for the SketchException class but cannot find it within the `processing.app`. We should keep a `SketchException` in `processing.app` and check if it works if it is just an extended version of the `processing.utils` one.

joshgiesbrecht:
I think we're good to go now. I wrote a new, very tiny processing.app.SketchException that extends the utils one, so not only does it allow other modes to work but it still plays nicely when testing if something is a processing.utils.SketchException.

Once this is released, I'll submit changes to the other modes so they refer directly to the utils class, and when *those* are released then I'll come back and delete the extra file. Hopefully. Fingers crossed. :laughing:

Stefterv:
Thank you @joshgiesbrecht! Not sure if the `super` wrappers are necessary, and could you mark the wrapper class as Deprecated?

joshgiesbrecht:
Deprecated marked. I didn't expect the wrappers would be necessary either, but Idea complained and said it was a compile problem, so I let it generate them. :shrug:

Stefterv (approved):

joshgiesbrecht:
Not sure how to connect this PR in github to the issue, but it's resolving issue https://github.com/processing/processing4/issues/1153

app/src/processing/app/SketchException.java
```diff
@@ -1,162 +1,30 @@
...
-  but WITHOUT ANY WARRANTY; without even the implied warranty of
-  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
-  GNU General Public License for more details.
-
-  You should have received a copy of the GNU General Public License
-  along with this program; if not, write to the Free Software Foundation,
-  Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA
-*/
+// temporary band-aid class to support modes which are still looking here for the now-refactored class.
+//  - josh giesbrecht
```
catilac, line 2: Just curious, when would this band-aid be removed?

joshgiesbrecht: My plan was to submit PRs to the other supported modes to switch them over, once this is released. Then when all existing supported modes are fully switched over this could be deleted.

Stefterv: Considering we do not know which 3rd party modes/tools have used this class I would expect this stub to be there for quite a while.

catilac: @joshgiesbrecht ah! also i just noticed you had already said this :) could you add a brief comment to the file documenting your plan?

catilac (changes requested):
This looks good to me. Please document your plan a bit in the file since it'll be around for a bit, and I'll approve!

app/src/processing/app/SketchException.java
joshgiesbrecht: @catilac @Stefterv There aren't that many modes to check, but I hadn't considered tools (or libraries, yikes). Is there a 'usual' for deprecated classes that should be planned for? I can update the file with specifics then.

Stefterv: The thing is, most modes we also do not have a way currently to easily publish new versions, so I suspect this stub to be there for quite a while

catilac (approved):
Going to approve. We'll have to come up with a plan for communicating the change to other project maintainers.

PR #1223, opened 2025-09-03
Delete java/lsp directory

Stefterv:
Cleanup of unused gradle files, originally the intent was to decouple the `lsp` from Processing but in the end it was easier to add the `processing lsp` command since it was so intertwined with everything else from the app


catilac:
Once that failing test passes, I'll merge. I'm rerunning now, it looks like an issue with snap.

PR #1226, opened 2025-09-04
Add warning for default pixelDensity behavior

Stefterv:
Introduces a warning message when pixelDensity defaults to 2x to match the display's pixel density. The warning prompts users to explicitly set pixelDensity in settings() to avoid the message.

Screenshot 2025-09-04 at 15 44 55


SableRaf (approved):
Looks good to me!

Note: I edited the warning message to be slightly more helpful: 

`Warning: Processing now sets pixelDensity(2) by default on high-density screens. This may change how your sketch looks. To revert to the old behavior, set pixelDensity(1) in setup().`

PR #1197, opened 2025-08-05
Debugger continues to work despite Sketch name and folder name are not equal

awood0727:
### Resolves
This PR resolves #1190 

### Changes
* Modified line 211 in java/src/processing/mode/java/debug/LineBreakpoint.java to allow .pde files to return its filename 
* Inserted a new function in java/src/processing/mode/java/debug/Debugger.java to check and translate PDE files if they are mismatched with their folder.

### Tests
- [x] We ran ./gradlew test, built successfully.
- [x] Ran ./gradlew run, built IDE successfully.
- [x] Tested the debugger to notice any change in its behavior.

#### Before Implementation
image

#### After Implementation
1190 issue solution

Issue #1190, opened 2025-07-24
Debugger does not work if Sketch name and folder are not equal


Stefterv:
### Most appropriate sub-area of Processing 4?

PDE

### Processing version

4.4.4

### Operating system

Linux/macOS

### Bug description

https://discourse.processing.org/t/impossible-to-activate-debug-in-processing/46781/8

> The reason … the folder and PDE file was not sharing the same name. When I renamed my PDE script with the same content folder it makes it works …

### Steps to reproduce this

1. Create and save a new example sketch
2. Close the sketch
3. Rename the folder
4. Open sketch in Processing
5. Say keep so that sketch name and folder name are no longer in sync
6. Enable Debugger
7. Create a breakpoint
8. Observe breakpoint not working

### snippet

```processing

// Paste your code below :)

void setup() {

}

void draw() {

}
```

### Would you like to work on the issue?

No, I’m just reporting the issue

Stefterv:
Related to #486

kaybcodes:
Hi! @Stefterv, I am looking into this issue with @SilasVM, @npNSU, and @awood0727. We were able to successfully replicate this issue using the steps provided. The debugger indeed fails to activate breakpoints when the sketch filename and folder name are not the same. We're currently investigating the underlying cause and exploring potential solutions. We'll follow up here with updates as we make progress.

npNSU:
Hey @Stefterv, Our team has been diligently working on finding the source and solution of this bug, and we think it could be along the line 912 of file path: [app/src/processing/app/Sketch.java](https://github.com/processing/processing4/blob/main/app/src/processing/app/Sketch.java)

Line 912: `if (!newMainName.equals(newSketchName) && newFolder.exists())`

We believe that this line of code is too restrictive for the user and it should be modified to allow the user to be able to change these variables without errors. 

Another thing we found was within lines 2170-2175 of the [processing4/app/src/processing/app/ui/Editor.java](https://github.com/processing/processing4/blob/main/app/src/processing/app/ui/Editor.java) file path, where it highlights the toggleBreakpoint() and the isDebuggerEnabled() functions. 

```
public boolean isDebuggerEnabled() {
    return false;
  }


  public void toggleBreakpoint(int lineIndex) { }
```
 
We think this may be vital for us to note, as it directly points out the workflow within the UI. 

We’d appreciate your feedback on our current investigation!

Stefterv:
[...]

PR #1227, opened 2025-09-05
Updated fix for the debugger issue

awood0727:
### Improves
* Issue #1190 
* Pull Request #1197 

###Changes 
- We have taken out the unnecessary function in java/src/processing/mode/java/debug/Debugger.java 

### Tests
- [x] We ran ./gradlew test, built successfully.
- [x] Ran ./gradlew run, built IDE successfully.
- [x] Tested the debugger to notice any change in its behavior.


Stefterv (approved):

Stefterv:
Tested on macOS and works as intended

catilac (approved):
Thank you! I think the `className` method can be DRY'ed out, and possibly we could use something like Guava, but I can propose that outside this PR.

PR #1230, opened 2025-09-06
ANGLE in Processing

TeoJT:
Google ANGLE is a compatibility layer which translates OpenGL commands into modern APIs like Direct3D, Vulkan, and Metal.
This PR is an attempt to add ANGLE to Processing as part of efforts to move away from OpenGL and JOGL (https://github.com/processing/processing4/issues/881)
GLFW is used as the windowing system, LWJGL used as a binding library to use OpenGL ES and GLFW.

## Current state
Window launches with basic animation using background(), but cannot actually draw anything on to the screen due to a pointer issue. No OpenGL errors detected and runs without any warnings.

## TODO:
- Fix issue where null pointer is passed in functions like glDrawElements() which causes nothing to render. 
- GLFW keyboard and mouse input
- ANGLE libraries (libEGL.dll, libGLESv2.dll, vulkan-1.dll) to be exported with sketch in sketch working dir.
- more...


catilac:
Hi @TeoJT thanks so much for this. I'm going to review before the rendering WG. Excited to see you there!

catilac:
Currently, this only runs on windows is that true? I would love to know what your approach in all these changes are at a higher level, and which files I should look at first so I can get a clearer understanding of where you've landed so far! 

Thank you!

TeoJT:
@catilac Yes, only have windows so far but I'll try to compile ANGLE for Linux too, I'll need help with MacOS as I don't have a Macbook.

For the high level view... I just created new renderers called P2DANGLE/P3DANGLE, they extend P2D/P3D by replacing JOGL with GLFW and LWJGL OpenGL ES, since GLFW supports ANGLE. The heart of it is PGLANGLE. PSurfaceANGLE is where the JOGL windows/inputs are replaced with GLFW. I want to see if it's possible to remove both native OpenGL and JOGL.

Only complication is that the ANGLE files (libEGL.dll, libGLESv2.dll, vulkan-1.dll) must be at the root dir of the sketch with no easy way to change where it's read from.

catilac:
@TeoJT Ahh thank you so much for that! I'll look keeping that in mind.

The complication you mentioned in regards to the DLLs needing to be in the root dir of sketch is definitely not great! Glad you noticed that. If we want to consider ANGLE we'll need to find an a way to change where it's read from in a maintainable way.

Issue #1219, opened 2025-08-28
`processing cli` does not support sketches with a different `main.pde` file


Stefterv:
### Most appropriate sub-area of Processing 4?

PDE

### Processing version

4.4.6

### Operating system

macOS

### Bug description

When running the Processing cli, having a sketch with a different main file, e.g. `sketch/main.pde` the cli returns that the sketch is not a valid sketch folder, `sketch/sketch.pde` does not exist. 

### Steps to reproduce this

1. Create a new sketch

2. Rename the main file 

3. Try to run the sketch with the cli

### Would you like to work on the issue?

No, I’m just reporting the issue

stepperpig:
Hi @Stefterv, I'd be interested in looking into this issue.

Stefterv:
Hi @stepperpig Great that you are interested, could you give us a quick overview of what your approach would be?

skools-here:
Hi, I’d like to work on this issue.

According to me currently, the CLI enforces the rule that the main .pde file must match the sketch folder name (e.g. sketch/sketch.pde). If the file is renamed (e.g. main.pde), it fails.

Instead, it must:

First look for a .pde file matching the folder name (current behavior).

If not found, fall back to the first .pde file in the sketch folder (same behavior the PDE GUI uses).

Add tests to ensure sketches with main.pde or other names can still run from the CLI.

I think this will work good.

Stefterv:
Hi @skools-here thank you for this description! I can tell you that the PDE looks for the sketch defined in the `sketch.properties` file if no matching PDE is found, if no sketch.properties exists then the process should error out.

skools-here:
Thanks for the clarification! 

So for the fix, I can update the CLI so that:

It first looks for <folderName>.pde as it does now.

If that’s missing, it will check sketch.properties for the main entry and use that file as the entry point.

If neither <folderName>.pde nor sketch.properties exists, the CLI will return an error.

This should bring the CLI in line with the PDE’s behavior. I’ll also add/update tests to cover the cases with main.pde and with/without sketch.properties.

Mr-KRAMA:
Hi @Stefterv, I'm interested in contributing to Processing. I see #1219 is already assigned, but I'd be happy to help with other CLI-related issues or review/test the current fix. Are there any other good first issues I could work on?

avinxshKD:
hey @Stefterv is this issue still available? I would like to work on this, pls assign

PR #1235, opened 2025-09-08
CLI now supports renamed main sketch file via sketch.properties .

skools-here:
This PR fixes issue [#1219]

Where the Processing CLI failed to run sketches when the main .pde file had been renamed (e.g. main.pde) and specified in sketch.properties.

Changes Made

Updated findMain() in Sketch.java:

If sketch.properties defines a main file and it exists → return it immediately.

If it does not exist → log a warning and fall back to the default <folderName>.pde.

If neither exists → return null (current error behavior preserved).

### DIFF
--- a/app/src/processing/app/Sketch.java
+++ b/app/src/processing/app/Sketch.java
@@ -1320,10 +1320,13 @@ public class Sketch {
       String main = props.get("main");
       if (main != null) {
         File mainFile = new File(folder, main);
-        if (!mainFile.exists()) {
+        if (mainFile.exists()) {
+          return mainFile;
+        }
+        else {
           System.err.println(main + " does not exist inside " + folder);
           // Fall through to the code below in case we can recover.
-          // Not removing the bad entry since this is a find() method.
+
         }
       }
     } catch (IOException e) {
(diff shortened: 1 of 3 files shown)


Stefterv:
Hi @skools-here, thank you for your efforts! I just tested it locally and the issue is still there. Could you give it a try on your side and update the fix?

Stefterv:
As part of your work could you please also add a test to https://github.com/processing/processing4/blob/main/app/test/processing/app/CLITest.kt

skools-here:
Ok i am looking into this and will also add test too !

skools-here:
Hey I fixed some issues in this kindly review !

Stefterv:
Hi @skools-here thank you for your effort, I quickly tested it again but still the same results, also the test does not actually test for the intended behaviour although it is heading in the right direction .
I can tell you that the offending code lives here: https://github.com/processing/processing4/blob/275d789c6480f97d0aa66ed53f6a747843ec298f/java/src/processing/mode/java/Commander.java#L148-L152

Stefterv:
Hi @skools-here are you still working on this?

skools-here:
No @Stefterv you can work on this .

hxrshxz:
@Stefterv Since no one's working on it Can I work on this? now, thank you

PR #1236, opened 2025-09-10
Fix the issue with the pre-processor test

Stefterv:
Upgrading the preprocessor made it no longer recognise the `quicktime` keyword in the `bug1532.pde` causing the test to fail. As far as I can see this test was never intended to test for that situation but rather for some invalid syntax later on line 50 of that file.

I cannot find anything about the `quicktime` keyword so I have just commented out the line with the new issue

### DIFF
--- a/java/test/resources/bug1532.pde
+++ b/java/test/resources/bug1532.pde
@@ -19,13 +19,6 @@ Capture cam;
 flatCube[][] grid;
 
 void setup() {
-  try {
-    quicktime.QTSession.open();
-  } 
-  catch (quicktime.QTException qte) { 
-    qte.printStackTrace();
-  }
-
   size (dx,dy,OPENGL);
   int d=day();
   int m=month();
--- a/java/test/processing/mode/java/ParserTests.java
+++ b/java/test/processing/mode/java/ParserTests.java
@@ -280,7 +280,7 @@ public class ParserTests {
 
   @Test
   public void bug1532() {
-    expectRecognitionException("bug1532", 50);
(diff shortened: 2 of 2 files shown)


Stefterv:
@SableRaf and I should do some archeological research into bug 1532 on the bugzilla archives

SableRaf:
@Stefterv Bug 1532 can be found at https://bugzilla.processing.org/1532

java/test/resources/bug1532.pde
```diff
@@ -20,7 +20,7 @@ flatCube[][] grid;
 
 void setup() {
   try {
-    quicktime.QTSession.open();
+    // quicktime.QTSession.open();
```
catilac: Is there a reason for leaving the commented out code in? Or would it make sense to remove the whole `try { ... }` block?

Stefterv: Nope, changed thanks!

Stefterv: Oh yes there is, the line-count is load bearing

catilac: @Stefterv oh is this because of a brittle test? or what do you mean?

Stefterv: The test is looking for an error occurring on line 50, so by removing these lines it fails, an alternative would be to update the line count I guess

catilac (changes requested):
Thank you! Looks good, please remove the unused code and update the test

catilac (approved):
Thanks @Stefterv!

PR #1239, opened 2025-09-11
Add 'Develop' menu for debug mode with update check

Stefterv:
Introduces a new 'Develop' menu that appears when debug mode is enabled. I realised I needed a quick way to trigger methods within the PDE which sometimes have no link to UI yet anywhere.

To enable the debug mode, click 5 times on the version number. 

Another nice to have that this PR adds is that there is a visual indicator on whether or not debug mode is active.

### DIFF
--- a/app/src/processing/app/ui/Editor.java
+++ b/app/src/processing/app/ui/Editor.java
@@ -48,19 +48,8 @@ import javax.swing.text.*;
 import javax.swing.text.html.*;
 import javax.swing.undo.*;
 
-import com.formdev.flatlaf.FlatLaf;
 import com.formdev.flatlaf.util.SystemInfo;
-import processing.app.Base;
-import processing.app.Formatter;
-import processing.app.Language;
-import processing.app.Messages;
-import processing.app.Mode;
-import processing.app.Platform;
-import processing.app.Preferences;
-import processing.app.Problem;
-import processing.app.RunnerListener;
-import processing.app.Sketch;
-import processing.app.SketchCode;
+import processing.app.*;
 import processing.utils.SketchException;
 import processing.app.contrib.ContributionManager;
 import processing.app.laf.PdeMenuItemUI;
@@ -147,6 +136,7 @@ public abstract class Editor extends JFrame implements RunnerListener {
   private FindReplace find;
   JMenu toolsMenu;
   JMenu modePopup;
+  JMenu developMenu;
 
   protected List<Problem> problems = Collections.emptyList();
 
@@ -680,6 +670,7 @@ public abstract class Editor extends JFrame implements RunnerListener {
       helpMenu.setText(helpMenu.getText() + " ");
     }
     menubar.add(helpMenu);
+    updateDevelopMenu(menubar);
 
     Toolkit.setMenuMnemonics(menubar);
     setJMenuBar(menubar);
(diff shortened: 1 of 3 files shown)


app/src/processing/app/ui/Editor.java
```diff
@@ -1060,6 +1051,37 @@ public JMenu buildModeMenu() {
 
   abstract public JMenu buildHelpMenu();
 
+  public void buildDevelopMenu(){
+    developMenu = new JMenu(Language.text("menu.develop"));
+
+    var updateTrigger = new JMenuItem(Language.text("menu.develop.check_for_updates"));
```
catilac, line 1057: I think this string needs to be "Check for updates"

catilac (commented):
I think the copy for the menu item needs to change, but looks good otherwise!
My questions since I'm learning still, how do I test the 5 clicks? The menu was just available for me.

app/src/processing/app/ui/Editor.java
Stefterv: If we want this menu item to have a readable name we need to add it to the language files, which I didn't want to pollute with a key that we might delete in the future

Stefterv:
The default run configuration will start Processing in DEBUG mode by setting the DEBUG environment variable to true.

Screenshot 2025-09-12 at 21 45 36

You can toggle DEBUG mode by clicking on the version number on the lower right of the editor 
Screenshot 2025-09-12 at 21 48 13
(unspecified in this image)

That will remove the `develop` menu from the menu bar as well.

app/src/processing/app/ui/Editor.java
catilac: Ah I see! Could you add a comment to explain?

app/src/processing/app/ui/Editor.java
```diff
@@ -1060,6 +1051,37 @@ public JMenu buildModeMenu() {
...
+  }
+
+  void updateDevelopMenu(JMenuBar menu){
+      if(menu == null){
+          menu = getJMenuBar();
+      }
+      if(developMenu == null){
+          buildDevelopMenu();
+      }
+      if(Base.DEBUG){
```
catilac, line 1077: this needs to be `if (Base.DEBUG == true)` I had set it to false in the run config , and the menu still showed up.

catilac (changes requested):
I left a comment about the conditional that checks Base.DEBUG, and I'm wondering if there is a way to trigger a redraw in the UI since there is a big delay when adding/removing the menu.

app/src/processing/app/ui/Editor.java
Stefterv: In Java a boolean is always a true or a false, so `if (Base.DEBUG == true)` specificity does not catch other states like it might in JS.

https://github.com/processing/processing4/blob/bdbd0254c75e7b809bf649c5042c9b7e5dd98bee/app/src/processing/app/Base.java#L68

Looking at `Base.DEBUG` setting it to false would still result in it I being true as it just checks if the key exists in the evn

Stefterv:
Oh for me the change is instant on macOS

app/src/processing/app/ui/Editor.java
Stefterv: I added it to the language file instead, also made it more descriptive

app/src/processing/app/ui/Editor.java
catilac: Oh! Just created an issue #1245 
Thanks for pointing that out!

catilac (approved):

PR #1192, opened 2025-07-28
Migration of the `dxf` library from Ant to Gradle

npNSU:
### Resolves
This PR resolves #981 

### Changes
* Replicated the code from Pull Request [#1189](https://github.com/processing/processing4/pull/1189)
* Edited the files: java/libraries/dxf/build.gradle.kts & java/build.gradle.kts
* Took out the dependency lines(17-21) in /java/libraries/dxf/build.gradle.kts that appeared in [PR 1189](https://github.com/processing/processing4/pull/1189/files).
* Modified line 89: `val libraries = arrayOf(“dxf")` in java/build.gradle.kts

### Tests
 * [x]  We ran `./gradlew test`, built successfully.
 * [x]  Ran `./gradlew run`, built IDE successfully.
 * [x]  Tested the creation of a .dxf file from a sketch using the following the sketch code: /Users/nia/processing4/java/libraries/dxf/examples/SimpleExport/SimpleExport.pde
 * [x]  Tested opening outputted .dxf file in [ShareCAD](https://sharecad.org/); displayed as expected

### Screenshots
Opened IDE after changes with test code and sketch
IDE Image

Produced .dxf file opened in ShareCAD
image

Thank you for your guidance @Stefterv! Please let us know if we need to make further adjustments!

PR #1240, opened 2025-09-11
Migration of the `dxf` library to Gradle

catilac:
Part of removing last bits of ant. Did I do this right? :)


Stefterv:
Looks good to me :)

Although I did find an unrelated bug by trying to run the example. I sense it has to do with the pre-processor migration so I will either create an issue or fix it today.

Le bug: #1241

catilac:
This has already been addressed by #1192

PR #1257, opened 2025-09-29
Use `start` instead of 0 for the starting index in `text`

iTutFadU:
`start` can be non-zero since the method is public. Otherwise it counts newlines before the start of the actual text in range.

### DIFF
--- a/core/src/processing/core/PGraphics.java
+++ b/core/src/processing/core/PGraphics.java
@@ -4768,7 +4768,7 @@ public class PGraphics extends PImage implements PConstants {
     }
 
 //    int start = 0;
-    int index = 0;
+    int index = start;
     while (index < stop) { //length) {
       if (chars[index] == '\n') {
         textLineAlignImpl(chars, start, index, x, y);


Stefterv:
Hi @iTutFadU could you show through screenshots what this change would entail?

iTutFadU:
The text will show up at the correct height (here green) instead of counting newlines preceding it before the range (here red).



The example code:
```processing
String text = "Some\n\ntext\nwith\nnewlines";
char[] chars;
float textWidth;

void setup() {
    size(500, 500, P2D);
    textFont(createFont("Arial", 50));
    textAlign(LEFT, TOP);
    chars = text.toCharArray();
    textWidth = textWidth(text.substring(6, text.length()));
}

void draw() {
    background(20);
    
    fill(#FF5555);
    text(chars, 6, text.length(), 0, 0);
    
    fill(#55FF55);
    text(text.substring(6, text.length()), width / 2, 0);
}
```

Stefterv (approved):

Stefterv:
@catilac

PR #1266, opened 2025-10-11
Fix signing property check in build script

Stefterv:
Corrects the property name from 'signing.signingInMemoryKey' to 'signingInMemoryKey' in the signing condition to ensure proper detection of signing configuration.

Verified with a release on my personal repo https://github.com/Stefterv/processing4/releases/tag/processing-1423-4.4.124


catilac:
Good catch, thanks so much @Stefterv !

PR #1277, opened 2025-10-14
Composable Preferences rewrite

Stefterv:
As part of the Jetpack Compose UI, we would like to have a reactive component that we can use to read and update Processing's preferences live. This PR will improve on the existing functionality and adds testing.

After #1276 as these file depend on the old implementation.

Part of #1280


catilac (approved):

catilac:
yeyyy : )

PR #1278, opened 2025-10-14
Rename output JARs in createLibrary tasks

Stefterv:
The createLibrary tasks in dxf, pdf, and serial modules now rename the output JAR files to dxf.jar, pdf.jar, and serial.jar respectively as Processing does not expect the version name to be in the library jar


catilac (approved):
lgtm! everything is versioned naturally by being in the java package.

Issue #1280, opened 2025-10-14
New Welcome Screen


catilac:
### Most relevant area for this enhancement?

PDE

### Feature to enhance

Welcome Screen in PDE

### Enhancement details

Build the New Welcome screen based on designs from Raph
Start UI Component Library based on those designs

### Would you like to help implement this enhancement?

No, I’m just suggesting the enhancement

catilac:
@Stefterv would you be able to map out what else needs to be done for the welcome screen?

Stefterv:
Sure, so far the list includes:

- [x] Find a reliable way for Jetpack Compose windows to size based on their contents
- [x] Wait for the design from @SableRaf 
- [x] Implement the design

catilac:
@Stefterv this is ready to be merged? It looked like the designs were completed and implemented?

Stefterv:
Yep! Although I need to finish the Settings.kt and Messages.kt before the whole branch can be merged

catilac:
@Stefterv would you mind tagging the issues for what else needs to be done from a technical standpoint on this task?

Stefterv:
Hi @catilac! Sure I just looked over the last PR again and as far as I see it is done

Stefterv:
I had mentioned still adding closing the window on escape & tests in the main body but that shouldn't be blocking. How do you feel about UI tests for the welcome screen at this stage?

catilac:
@Stefterv UI tests sound good if you feel like the functionality feels stable enough to you. We also have to discuss how to roll out this welcome screen so that gives us some time and space.

catilac:
I mentioned this in the code review, but going to put it here for brevity: we do need to figure out the copy for the welcome screen

PR #1283, opened 2025-10-15
Refactor Locale class and add LocaleProvider test

Stefterv:
As part of the Jetpack Compose UI, we would like to have a reactive component that we can use to read and set Processing's Locale live. This PR will improve on the existing functionality and adds testing.

Part of #1280

### DIFF
--- a/app/src/processing/app/ui/theme/Locale.kt
+++ b/app/src/processing/app/ui/theme/Locale.kt
@@ -21,7 +21,7 @@ import java.util.*
  * val localizedString = locale["someKey"]
  * ```
  */
-class Locale(language: String = "", val setLocale: (java.util.Locale) -> Unit) : Properties() {
+class Locale(language: String = "", val setLocale: ((java.util.Locale) -> Unit)? = null) : Properties() {
     var locale: java.util.Locale = java.util.Locale.getDefault()
 
     init {
@@ -46,7 +46,7 @@ class Locale(language: String = "", val setLocale: (java.util.Locale) -> Unit) :
     }
     operator fun get(key: String): String = getProperty(key, key)
     fun set(locale: java.util.Locale) {
-        setLocale(locale)
+        setLocale?.invoke(locale)
     }
 }
 /**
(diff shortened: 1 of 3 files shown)


app/src/processing/app/ui/theme/Locale.kt
```diff
@@ -1,24 +1,41 @@
...
     init {
-        val locale = java.util.Locale.getDefault()
-        load(ClassLoader.getSystemResourceAsStream("PDE.properties"))
-        load(ClassLoader.getSystemResourceAsStream("PDE_${locale.language}.properties") ?: InputStream.nullInputStream())
-        load(ClassLoader.getSystemResourceAsStream("PDE_${locale.toLanguageTag()}.properties") ?: InputStream.nullInputStream())
-        load(ClassLoader.getSystemResourceAsStream("PDE_${language}.properties") ?: InputStream.nullInputStream())
+        loadResourceUTF8("PDE.properties")
+        loadResourceUTF8("PDE_${locale.language}.properties")
+        loadResourceUTF8("PDE_${locale.toLanguageTag()}.properties")
+        loadResourceUTF8("PDE_${language}.properties")
```
catilac, line 31: this is so nice

app/src/processing/app/ui/theme/Locale.kt
```diff
@@ -28,18 +45,86 @@ class Locale(language: String = "") : Properties() {
...
+        else -> LayoutDirection.Ltr
+    }
 
-    val locale = Locale(languageFile.readText().substring(0, 2))
-    CompositionLocalProvider(LocalLocale provides locale) {
-        content()
+    CompositionLocalProvider(LocalLayoutDirection provides dir) {
+        CompositionLocalProvider(LocalLocale provides locale) {
+            content()
+        }
```
catilac, line 128: Ah cool! So this is like how things are composable? People can pass in a block that will be wrapped by this Locale work

catilac (approved):
looks good! i left a couple questions for my own learning and curiosity

app/src/processing/app/ui/theme/Locale.kt
Stefterv: Exactly, if you want access to localised strings you do 


    LocaleProvider {
      val locale = LocalLocale.current
      val localizedString = locale["someKey"]
    }

And the beatiful things about these contexts is that if you want to display say something in Spanish within the English interface, you can override the `LocalLocale` with a `CompositionLocalProvider`. In that case its not super helpful but there's a lot of other cases where its very much is

Issue #1269, opened 2025-10-12
Upgrade JDK


catilac:
See if we can upgrade project to use latest JDK 25 LTS
Or at least JDK 22+ to allow us to use Project Panama

tychedelia:
Seems like there's some native loading issues in 25 that are causing the PDE to be a bit wonky. Sketches run fine.

<img width="1624" height="1786" alt="Image" src="https://github.com/user-attachments/assets/897528a4-a1d9-414b-b7ff-7cfb1b9b2cf3" />

Stefterv:
Right I have seen this before when working on the pre-processor

catilac:
I'll make a task for updating the pre-processor just to track it. But i'm glad the sketches are building. I wonder what test infrastructure we could get into place to track that as we go

WilliamSenn1967:
Hello there...

Is it possible to call the settings function which is filled with some commands that were filtered out by the preprocessor?
You can locate the settings function near the main function... the settings function is never called, but it is there... I think the settings function is always at the bottom of any sketch compiled / preprocessed to java file...

with kind regards,

William Senn.

P.S. My software is depending on fullScreen(), pixelDensity(1), smooth()  which are now never executed because the settings function is never called, but only preprocessed....

catilac:
Hi @WilliamSenn1967! Thank you for testing this!

glvjr:
Hello,

I am using this library successfully with Processing:
https://fazecast.github.io/jSerialComm/

Please keep issues (opportunities) like this in mind for future:

> Starting with Java version 24 and later, additional restrictions have been placed on applications that make calls to native code. To allow your Java application to use the jSerialComm library, you may need to pass one of the following command-line flags to your application invocation:
> 
> --enable-native-access=com.fazecast.jSerialComm: to allow native code access only to jSerialComm
> --enable-native-access=ALL-UNNAMED: to allow native code access to all libraries used by your app.

Will this impact other libraries that make calls to native code?

I could always make a custom build with some guidance on where to modify this.

Thank you in advance!

PR #1287, opened 2025-10-15
Initial FFI setup

### LINKED ISSUE #1270: Initial Build Setup For Renderer
Make sure we can build project using Gradle.
Check for existing Gradle/Cargo integration, otherwise implement our own

Create basic Rust FFI library, and see that it can be called on every platform using Project Panama.

tychedelia:
Closes #1270
Partially addresses #1269 

TODO:
- [x] Add docs on how to install jextract
- [x] Remove Kotlin EAP
- [x] Add README to include folder
- [x] Add jextract to build
- [x] Wire `cargo clean` into gradle.


app/build.gradle.kts
```diff
@@ -1,15 +1,94 @@
...
 import org.jetbrains.compose.desktop.application.tasks.AbstractJPackageTask
 import org.jetbrains.compose.internal.de.undercouch.gradle.tasks.download.Download
 import java.io.FileOutputStream
 import java.util.zip.ZipEntry
 import java.util.zip.ZipOutputStream
+import javax.inject.Inject
 
 // TODO: Update to 2.10.20 and add hot-reloading: https://github.com/JetBrains/compose-hot-reload
 
+abstract class SignResourcesTask : DefaultTask() {
```
Stefterv: Good one! Is the intention to move this to a separate file?

build.gradle.kts
```diff
@@ -12,6 +12,19 @@ plugins {
 // Can be deleted after the migration to Gradle is complete
 layout.buildDirectory = file(".build")
 
+allprojects {
```
tychedelia, line 17: This is just a workaround until kotlin `2.3.0` is released with jdk 25 support.

app/build.gradle.kts
tychedelia: i think it probably should be in a separate file, totally. will do that....

build.gradle.kts
catilac: can we leave a little TODO saying that?

core/src/processing/core/NativeLibrary.java
```diff
@@ -0,0 +1,110 @@
...
+    private static final String OS_NAME = System.getProperty("os.name").toLowerCase();
+    private static final String OS_ARCH = System.getProperty("os.arch").toLowerCase();
+
+    private static final String platform;
+    private static final String architecture;
+    private static final String libraryExtension;
+
+    static {
+        // platform
+        if (OS_NAME.contains("mac") || OS_NAME.contains("darwin")) {
```
Stefterv, line 25: Does it make sense to start including Android from the start?

libProcessing/ffi/include/processing.h
```diff
@@ -0,0 +1,17 @@
+#ifndef PROCESSING_H
+#define PROCESSING_H
+
+#pragma once
+
+/* Generated with cbindgen:0.29.0 */
+
+/* Warning, this file is autogenerated by cbindgen. Don't modify this manually. */
```
Stefterv: If this is generated is there a good reason to leave it out of the `.gitignore`?

core/build.gradle.kts
```diff
@@ -35,6 +35,138 @@ dependencies {
     testImplementation(libs.junit)
 }
 
+val osName = System.getProperty("os.name").lowercase()
```
Stefterv: Does running the `clean` gradle task also clean the Rust files?

libProcessing/ffi/include/processing.h
tychedelia: it's potentially useful to be able to review the header diffs just to see what the ffi interface looks like. this is what tools will ultimately bind against, and while it's mostly easy to translate mentally 1:1 from the rust source, i think it can be nice to have the header in source control, but i could go either way.

Stefterv: Okay let's put that argument in a readme.md in that folder

core/src/processing/core/NativeLibrary.java
tychedelia: ugh yeah, good callout. i did identify that there's a plugin for building Rust/android here https://github.com/mozilla/rust-android-gradle. let me think more about this re: the binding strategy, since panama won't work there

core/build.gradle.kts
tychedelia: good q: no it doesn't right now, although probably should? ime clean is a little less common in rust than java as it leads to quite expensive rebuilds but it should probably be wired in to be consistent for the clean rebuild cycle. i'll note this as a task

core/src/processing/core/NativeLibrary.java
catilac: Whatever you do, I don't think android should block progress this early on.

libProcessing/ffi/Cargo.toml
```diff
@@ -3,4 +3,12 @@ name = "ffi"
 version = "0.1.0"
 edition = "2024"
 
+[lib]
+name = "processing"
+crate-type = ["cdylib"]
```
catilac, line 8: obsessed with what we get from rust

catilac (commented):
Left some comments! How would I try this? 
I'm happy to approve and get it into our feature branch, but the TODOs tell me this is just early feedback (which I love)

Exciting work!

libProcessing/ffi/include/processing.h
tychedelia: I reversed myself here and added it to gitignore.

core/build.gradle.kts
tychedelia: done

core/src/processing/core/NativeLibrary.java
tychedelia: I added a `enableWebGPU` flag that enforces source level compat with android atm. we can make the call on full android support later

app/build.gradle.kts
tychedelia: factored out all these into a new `buildSrc` dir

build.gradle.kts
tychedelia: it's more complicated now as it enforces android source level compat

PR #1293, opened 2025-10-16
libprocessing ffi error handling

tychedelia:
Sets up basic bevy `App` initialization and cross-FFI error handling.

Our strategy here is the following: Java should do any validation necessary before calling into the FFI layer. Errors are assumed to be unrecoverable and evidence of us doing something wrong. So we don't return error codes or anything but just check the error status as a string so we can at least throw a helpful exception for bug reports.

### DIFF
--- a/core/src/processing/webgpu/PWebGPU.java
+++ b/core/src/processing/webgpu/PWebGPU.java
@@ -1,7 +1,12 @@
 package processing.webgpu;
 
 import processing.core.NativeLibrary;
-import processing.ffi.processing_h;
+
+import java.lang.foreign.MemorySegment;
+
+import static java.lang.foreign.MemorySegment.NULL;
+import static processing.ffi.processing_h.processing_check_error;
+import static processing.ffi.processing_h.processing_init;
 
 /**
  * PWebGPU provides the native interface layer for libProcessing's WebGPU support.
@@ -9,7 +14,8 @@ import processing.ffi.processing_h;
 public class PWebGPU {
 
     static {
-        NativeLibrary.ensureLoaded();
+        ensureLoaded();
+        init();
     }
 
     /**
@@ -20,9 +26,25 @@ public class PWebGPU {
     }
 
     /**
-     * It's just math, silly!
+     * Initializes the WebGPU subsystem. Must be called before any other WebGPU methods.
      */
-    public static long add(long left, long right) {
-        return processing_h.processing_add(left, right);
+    public static void init() {
+        processing_init();
+        checkError();
+    }
+
+    /**
+     * Checks for errors from the native library and throws a PWebGPUException if an error occurred.
+     */
+    private static void checkError() {
+        MemorySegment ret = processing_check_error();
+        if (ret.equals(NULL)) {
+            return;
+        }
+
+        String errorMsg = ret.getString(0);
+        if (errorMsg != null && !errorMsg.isEmpty()) {
+            throw new PWebGPUException(errorMsg);
+        }
     }
 }
--- a/libProcessing/Cargo.toml
+++ b/libProcessing/Cargo.toml
@@ -1,8 +1,3 @@
-[package]
-name = "libProcessing"
-version = "0.1.0"
-edition = "2024"
-
 [workspace]
 resolver = "3"
 members = ["ffi","renderer"]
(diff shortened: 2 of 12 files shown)


libProcessing/Cargo.toml
```diff
@@ -1,8 +1,3 @@
-[package]
```
catilac, line 1: curious about why this was removed? is it something to do with cargo workspaces?

libProcessing/renderer/Cargo.toml
```diff
@@ -4,3 +4,9 @@ version = "0.1.0"
 edition = "2024"
 
 [dependencies]
+tracing = "0.1"
+tracing-subscriber = "0.3"
+bevy = { version = "0.16", no-default-features = true, features = [
+    "bevy_render"
```
catilac, line 10: for wayland windowing we need to add `"wayland"` to the features array

libProcessing/Cargo.toml
tychedelia: ah! yeah, so this just makes it a "virtual" workspace, i.e. with no top level bin or lib output. i think in the future we'd want to use a top level crate for some of the following reasons: we want to promote ffi to the "main" focus, we want to produce a `processing` rust lib, we want to use it as a bin target for python, etc.

libProcessing/renderer/Cargo.toml
catilac: if it's helpful, i can do it separately so i can actually test

tychedelia: we'll def need more fleshing out here, i think there's a bunch of other stuff we'll want too. we're just generally going to be doing the "annoying" bevy integration version where we fine grain select what we want. this is to keep the final dylib small for packaging. see also the recently merged https://github.com/bevyengine/bevy/pull/21472 which could be helpful for us.

libProcessing/Cargo.toml
catilac: Ah, ok! Thank you!

libProcessing/renderer/Cargo.toml
catilac: that looks cool. maybe useful for per-target customization

catilac (approved):
Great! I like the error approach too. And if we ever for whatever reason needed to share more details then we can just serialize more information.

I'm approving, but also curious about what things will look like, say, if they write an erroneous compute shader.

core/src/processing/webgpu/PWebGPU.java
```diff
@@ -20,9 +26,25 @@ public static void ensureLoaded() {
...
+        processing_init();
+        checkError();
+    }
+
+    /**
+     * Checks for errors from the native library and throws a PWebGPUException if an error occurred.
      */
-    public static long add(long left, long right) {
-        return processing_h.processing_add(left, right);
+    private static void checkError() {
```
Stefterv, line 39: So this means we would need to poll the rust side for any errors?

tychedelia:
> I'm approving, but also curious about what things will look like, say, if they write an erroneous compute shader.

i think we'll def run into situations like this. what i'd propose is that we write a `validate_shader` ffi function that can eagerly check and return structured output. but also, there's nothing preventing us from returning a custom ffi error type for a given operation where it would be helpful, for example:

```rust
struct ValidationResult {
    error_code: u8,
    data: *const c_void,
}
```

in rust applications, you typically will bubble up `Result` using `?` to a top level handler that will either log or panic. and so the pattern in this pr basically just ensure that we can reliably return results from our internal library code.

on the java side, we're using unchecked exceptions. right now it's just a general "something went wrong" exception class, but we could totally add an error code in addition to message and sub-type more on the java side if we do want to throw something like `ValidationException`.

core/src/processing/webgpu/PWebGPU.java
tychedelia: yup, this is a tedious consequence of ffi code. there are a few different c-style error patterns you can use, but they're all pretty manual. rust exceptions (panics) can't cross the ffi boundary (doing so is undefined behavior).

realistically, we'll probably only end up having like 20ish ffi functions. if we had hundreds, i'd suggest writing an annotation processor or some other kind of reflection magic in java to automate this.

Stefterv: Okay and there is no way from Rust to push an event into Java?

PR #1297, opened 2025-10-19
Add diagnostics gathering feature to debugger and UI

### LINKED ISSUE #1294: Gathering System Diagnostics from a sketch over the debug connection

PDE

### Feature to enhance

Diagnostics

#### Enhancement description
Previously I added the ability to copy system information to the clipboard by clicking on the version number. 
Inspired by Glv’s post https://discourse.processing.org/t/report-system-and-environment-diagnostics-at-runtime/47384
I realised that we could use the debug connection to a sketch to gather even more data about the system. 

#### Benefits
Increasing the information gathered will make it easier to find and fix bugs in the future. Especially if we start asking users to submit these information dumps. 

#### Challenges
Privacy

#### Additional context
The way this would need to be done is by creating a data structure and function within ‘core’ that will gather the data that is not available in de PDE. Then the PDE needs to check if a debug connection is active in the ‘JavaEditor’ and call the function present in ‘core’. 

No, I’m just suggesting the enhancement

hxrshxz:
Fixes #1294
## Description 
- Adds ability to gather system diagnostics from a running sketch through the debugger.


Stefterv:
Hi @hxrshxz Thank you so much for this work! 

A few things: 
- Rather than communicating over JSON and writing a custom writer and parser, lets use the build-in communication in the Debug connection
- The debug connection is active even when the debugger is not connected in Java mode, so this is not limited to running a sketch with the Debugger active
- Rather than adding a whole new screen, let's just add it to the clipboard content that already exists when you click the version number

hxrshxz:
Hi @benfry, thank you for the feedback. I've updated the PR to address your points:

1. Direct JDWP Access: Switched to reading PApplet fields directly via JDI, removing the need for JSON or custom protocols.
2. Universal Diagnostics: Implemented a static Debugger.getDiagnostics(Runner) that works for both Run and Debug modes without requiring an active debug session.
3. Clipboard Integration: Simply appends sketch diagnostics to the existing version-click clipboard action. No new UI elements were added.

really sorry for taking so lonng 

https://github.com/user-attachments/assets/2cd933f0-4c18-46ca-b5c2-15ca7b2aa74d

hxrshxz:
Hey @Stefterv can you please have a look when you are free , thanks for your time

Stefterv (approved):

Stefterv:
Looks good @hxrshxz! Thank you!

Optionally, I think it would be good to add the same functionality to the new settings screen as the `Generate Report` button is much easier to find. 
Also I think me and @catilac would need to think what kind of debug information would be helpful (renderer for example)
But neither are blockers for this PR so that can also be separate issues.

catilac (approved):

PR #1298, opened 2025-10-21
Theming

Stefterv:
Introduces light and dark themes using Material3. Refactors Theme.kt to use Material3 theming, adds a PDETheme composable, and provides a desktop preview app for theme components. Updates Typography.kt to use Space Grotesk font family and defines new typography styles for Material3.

Screenshot 2025-10-21 at 10 51 37
Actual colours will still be updated

Related to #1280


catilac (approved):
looking good!

PR #1299, opened 2025-10-22
General cleanup of `Base`

Stefterv:
I started cleaning up some of `Base`'s startup sequence for clarity of what is being started when. Nowhere near completion and I think a lot of this class will need to be refactored in the future.

Also removed some of the timing measurement comments

Added some comments to the Processing CLI class

### DIFF
--- a/app/src/processing/app/Base.java
+++ b/app/src/processing/app/Base.java
@@ -51,11 +51,15 @@ import processing.data.StringList;
  * files and images, etc.) that comes from that.
  */
 public class Base {
-  // Added accessors for 0218 because the UpdateCheck class was not properly
-  // updating the values, due to javac inlining the static final values.
+  /**
+   * Revision number, used for update checks and contribution compatibility.
+   */
   static private final int REVISION = Integer.parseInt(System.getProperty("processing.revision", "1295"));
-  /** This might be replaced by main() if there's a lib/version.txt file. */
-  static private String VERSION_NAME = System.getProperty("processing.version", "1295"); //$NON-NLS-1$
+  /**
+   * This might be replaced by main() if there's a lib/version.txt file.
+   *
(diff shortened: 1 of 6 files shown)


app/src/processing/app/Base.java
```diff
@@ -484,55 +486,42 @@ static public boolean isCommandLine() {
...
     Platform.initBase(this);
 
-    long t6 = System.currentTimeMillis();
+    // check for updates
+    UpdateCheck.doCheck(this);
+
 
-//    // Check if there were previously opened sketches to be restored
-//    boolean opened = restoreSketches();
+    ContributionListing cl = ContributionListing.getInstance();
```
Stefterv, line 515: This is one thing that moved in the sequence, I think this should be okay

app/src/processing/app/Base.java
```diff
@@ -317,24 +244,51 @@ static private void createAndShowGUI(String[] args) {
...
+        Messages.showError("Trouble without a name",
+                           "Could not create a place to store untitled sketches.\n" +
+                           "That's gonna prevent us from continuing.", e);
     }
   }
 
+  private static void setLookAndFeel() {
+        try {
+            // Use native popups to avoid looking crappy on macOS
+            JPopupMenu.setDefaultLightWeightPopupEnabled(false);
```
Stefterv, line 276: Also moved in the sequence

app/src/processing/app/Base.java
catilac: what is contribution listing?

Stefterv: It's the helper class that will gather the contributions in the contributions manager from the net

catilac (approved):
lgtm! merge when ready

PR #1300, opened 2025-10-22
Refactor initialization to use background threads

Stefterv:
Moved several initialization tasks to background threads for improved startup performance and responsiveness, including tool, font, UI, and suggestion generator setup. Deferred some tool initialization until first use, added caching for SVG rendering, and simplified string loading in PApplet. Also made minor changes to library list and variable inspector initialization.

- [ ] Find a way to display the splash screen optionally as `Preferences` will not yet be initialised.

https://github.com/user-attachments/assets/6aec6398-0cfe-4628-8487-69c88f41bfc3

### DIFF
--- a/app/src/processing/app/Base.java
+++ b/app/src/processing/app/Base.java
@@ -23,6 +23,17 @@
 
 package processing.app;
 
+import java.awt.*;
+import java.awt.event.ActionListener;
+import java.io.*;
+import java.lang.reflect.InvocationTargetException;
+import java.util.*;
+import java.util.List;
+import java.util.Map.Entry;
+
+import javax.swing.*;
+import javax.swing.tree.DefaultMutableTreeNode;
+
 import com.formdev.flatlaf.FlatDarkLaf;
 import com.formdev.flatlaf.FlatLaf;
 import com.formdev.flatlaf.FlatLightLaf;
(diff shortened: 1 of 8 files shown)


catilac:
Cool! How much did performance improve? Also what is going on with the preferences not being initialized meaning adding a new settings thing.

Do any of these threads need to be joined at some point? Or it's fine for all of them to just start and end whenever?

Stefterv:
Performance went from 1500ms to 350ms startup time, so now the welcome & splash screen itself is actually the biggest bottleneck in starting. 

The `Preferences` class is not necessarily initialised when starting the Welcome screen, we could either initialise it early (but I think that might cause other issues) or we could use Java's `Preferences` maybe instead. 

As for joining the threads, I wanted to look into this further, so far in my testing the PDE starts just fine and I haven't noticed any issues, but that's only tested on one devices on one os, so not particularly extensive (:

tychedelia:
Have you considered using an `ExecutorService` instead of manually creating all the threads? That could provide better control or allow thread reuse particularly for tasks that are small.

Stefterv:
No I didn't know of that! I think in general the startup sequence needs a revisit, so this PR is a stopgap to at least get some speedup without doing a larger refactor as I don't see that happening anytime soon

catilac:
If there is a thread pool manager we can use to manage threads we should use that. I'd prefer we not introduce the use of low-level concurrency primitives on the app startup code paths to improve startup performance when it sounds like it all needs to be changed anyways.

PR #1302, opened 2025-10-22
WebGPU surface creation / basic app lifecycle

tychedelia:
Implements the basic rendering lifecycle needed to drive bevy forward from the processing event loop.

When a `PSurfaceGLFW` is created, we pass the window handle over ffi and import it into bevy. Bevy then will create the underlying `wgpu` surface via its ordinary operations for windows. The `Window` here we insert is a non-functional facade -- we'll make sure to synchronize it from the Java side but have no need atm to allow bi-directional communication (i.e. bevy resizes Glfw).

There's some additional work we'll need to do to think about multi-window support. Right now we're driving the wgpu renderer forward in `endDraw` on `PGraphicsWebGPU` but this may not be correct in the presence of multiple graphics contexts.

Basically, for implementing the main processing api, what we'll want to do is run all the user's functions to build up a frame's worth of abstract draw data on the Rust side (i.e. insert stuff into the ECS world as needed) and then run `update` once at the end of frame which will drive the world forward one tick and present to the surface. 

## GLFW + AWT = :(

GLFW does *not* play nice with AWT, especially on macOS where we have the requirement to run on the main thread. This is well documented including in this repo and will be an issue long term. Right now, I've chosen to just hard code disabling AWT when launching a sketch, but we'll need a more comprehensive solution here eventually.

## How to test

Right now, I've only added support for macOS because I don't have my other machines to test on. But if you have a mac it should be pretty easy, just add a basic bevy scene to the app init. The next PR should implement our first basic actual rendering to test from Java.

### DIFF
--- a/core/src/processing/webgpu/PSurfaceGLFW.java
+++ b/core/src/processing/webgpu/PSurfaceGLFW.java
@@ -1,0 +1,423 @@
+package processing.webgpu;
+
+import org.lwjgl.glfw.GLFW;
+import org.lwjgl.glfw.GLFWErrorCallback;
+import org.lwjgl.glfw.GLFWFramebufferSizeCallback;
+import org.lwjgl.glfw.GLFWNativeCocoa;
+import org.lwjgl.glfw.GLFWWindowPosCallback;
+import org.lwjgl.system.MemoryUtil;
+import org.lwjgl.system.Platform;
+
+import processing.core.PApplet;
+import processing.core.PGraphics;
+import processing.core.PImage;
+import processing.core.PSurface;
+
+import java.io.File;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.locks.Condition;
+import java.util.concurrent.locks.Lock;
+import java.util.concurrent.locks.ReentrantLock;
+
+/**
+ * GLFW-based surface implementation for WebGPU.
+ */
+public class PSurfaceGLFW implements PSurface {
+
+    protected PApplet sketch;
+    protected PGraphics graphics;
+
+    protected long window;
+    protected boolean running = false;
+
+    protected boolean paused;
+    private final Lock pauseLock = new ReentrantLock();
+    private final Condition pauseCondition = pauseLock.newCondition();
+
+    protected float frameRateTarget = 60;
+    protected long frameRatePeriod = 1000000000L / 60L;
+
+    private static final AtomicInteger windowCount = new AtomicInteger(0);
+    private static AtomicBoolean glfwInitialized = new AtomicBoolean(false);
+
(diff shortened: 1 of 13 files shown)


core/src/processing/webgpu/PSurfaceGLFW.java
```diff
@@ -0,0 +1,423 @@
...
+    @Override
+    public Object getNative() {
+        return window;
+    }
+
+    public long getWindowHandle() {
+        if (Platform.get() == Platform.MACOSX) {
+            return GLFWNativeCocoa.glfwGetCocoaWindow(window);
+        } else {
+            throw new UnsupportedOperationException("Window handle retrieval not implemented for this platform");
```
tychedelia, line 124: @catilac We'll need to add wayland/x11 detection here for linux. Window I think should be pretty easy I just can't test right now.

catilac: I'll get it going on wayland!

libProcessing/ffi/src/error.rs
```diff
@@ -34,16 +35,28 @@ pub fn clear_error() {
...
-    match result {
-        Ok(value) => {
-            Some(value)
-        }
+pub fn check<T, F>(f: F) -> Option<T>
+where
+    F: FnOnce() -> Result<T, ProcessingError> + panic::UnwindSafe,
+{
+    // we'll catch panics here to prevent unwinding across the FFI boundary
+    panic::catch_unwind(|| match f() {
```
catilac, line 44: ooh good catch. i'll be sure to pay attention to this

libProcessing/renderer/Cargo.toml
```diff
@@ -6,7 +6,12 @@ edition = "2024"
 [dependencies]
 tracing = "0.1"
 tracing-subscriber = "0.3"
-bevy = { version = "0.16", no-default-features = true, features = [
+bevy = { version = "0.17", no-default-features = true, features = [
```
catilac, line 9: YAY

libProcessing/renderer/Cargo.toml
```diff
@@ -6,7 +6,12 @@ edition = "2024"
...
     "bevy_render"
 ] }
-thiserror = "2"
\ No newline at end of file
+thiserror = "2"
+raw-window-handle = "0.6"
```
catilac, line 13: is this a library to help us manage a raw window handle to pass into JVM world?

catilac (commented):
Amazing! Merge when ready. I'll get started on wayland support since I'm running it.
Also I'm seeing how this lib can be consumed by other languages like C or python

libProcessing/renderer/Cargo.toml
tychedelia: yes! it's what's used by wgpu internally

PR #1308, opened 2025-10-27
Implement `background` method

tychedelia:
Allows setting the background clear color in a sketch.

Changes:
- Adds a camera when spawning a window. Note that we also add a `RenderLayers` component, which we'll also need to attach to any geometry spawned to render in that window. This is just hygiene for eventual multi-window apps.
- Adds a simple `Color` FFI struct. We'll likely want to make this more complicated later or support all the color spaces that bevy does (even if processing doesn't). But this demonstrates how to do allocations on the Java side.

I didn't implement the `PImage` override here as that will require the actual renderer (i.e. not just a clear color). It's also possible the semantics here are wrong. I think processing may allow you to clear multiple times in a frame, which means drawing a fullscreen quad not setting the render pass clear. But we can follow up with that once more rendering infra is implemented.

### DIFF
--- a/.github/workflows/build-gradle.yml
+++ b/.github/workflows/build-gradle.yml
@@ -17,6 +17,10 @@ jobs:
         with:
           java-version: '17'
           distribution: 'temurin'
+      - name: Install Wayland Dependencies
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y libwayland-dev libxkbcommon-dev
       - name: Setup Gradle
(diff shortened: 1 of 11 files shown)


catilac:
Hi! What is a good way to test this?

tychedelia:
> Hi! What is a good way to test this?

You can create a sketch that sets the background color (:

catilac (approved):
it works! and noted on the multiple clear calls.

PR #1313, opened 2025-11-04
Move contributor list to CONTRIBUTORS.md

SableRaf:
Created CONTRIBUTORS.md and updated .all-contributorsrc to reference the new file instead of README.md. This will reduce the size of the README and improve loading times.


catilac (approved):
Amazing! Great PR too.

SableRaf:
Thanks!

Note: A similar change on p5.js repo caused the bot to fail but @davepagurek figured out there needs to be something before the contributors table for the bot to recognize it. Since this PR does include a paragraph before the table we should be fine but it might be worth adding a header above just to be safe (say we forget about this issue and decide to remove or relocate that paragraph later).

https://github.com/processing/p5.js/pull/8233

SableRaf:
@all-contributors please add @Qianqianye for fundingFinding 

(This is a test)

PR #1314, opened 2025-11-05
Allow exported apps on windows to be run from anywhere on the system

jSdCool:
On Linux and Macos an exported sketch can be run from any working directory. Until now that has not been the case on windows. This was caused by the EXEDIR environment variable not being used when forming the class path. This pr makes that simple change, allowing the executable generated from exporting a sketch to be executed on the command line from any working directory.


Stefterv (approved):

Stefterv:
Hi @jSdCool Thanks for this quick fix, I tested it on Windows just now and indeed allows the sketch to be run from any working dir! Thanks!

catilac (approved):
lgtm!

Issue #1320, opened 2025-11-07
`loadPixels` and `PGraphics` dependency batching in WebGPU


tychedelia:
The processing API allows performing arbitrary GPU->CPU readback on demand via the `loadPixels` method. This is generally an extension of the immediate mode philosophy of processing but poses significant friction with modern graphics architectures. In contrast with opengl, which goes to significant lengths to preserve the illusion that GPU data is easily accessible, WebGPU is designed to foreground that GPU operations are asynchronous relative to the CPU timeline. While the API does support blocking on these operations, the expectation of most framework including `bevy` is that you won't.

Concretely why this matters is because modern graphics libraries really want to batch rendering together for efficiency reasons. For example, in Bevy, all cameras (which could be considered an analog to a `PGraphics` instance) want to render at the same time every frame. `loadPixels` introduces an architectural complication in that the current draw state may need to be flushed and made visible to any other graphics context at any arbitrary time.

In other words, beyond any performance concerns, this highlights a potential dependency problem between `PGraphics` instances: 
- In the event where multiple `PGraphics` instances aren't dependent on each other, batching works fine and we can delay flushing all their draw state til the end of frame. 
- When a `PGraphics` is used in another `PGraphics`, e.g. an off-screen texture used by something in rendering to the screen, we could simply just track a relative order to ensure the Bevy `Camera` for one runs before the other.
- If the user calls `loadPixels`, we need to flush the draw state right now and make the texture visible to potentially any other CPU code.

## Approach for WebGPU with Bevy

 We should start just by mirroring the immediate mode API. What this means is that we'll mirror the `drawStart`, `flush`, `drawEnd` lifecycle. Set `CameraOutputMode::Skip`at the beginning of each frame for each surface to render only to the intermediate texture and `CameraOutputMode::Write` only when calling `drawEnd`.

We can handle `loadPixels` in this way just by doing a synchronous readback after a `flush`. It's fine, and will superficially look like opengl's behavior from the user perspective.

Because this immediate mode approach likely has some undesirable overhead (although tbd, it may be pretty minimal), we can do our own dependency/dirty tracking if necessary down the line. What this would look like is keeping an in-flight dependency graph of `PGraphics` and how they're being used, and only trigger flushes when they *acutally* need to be made visible to other graphics contexts. This isn't hard to do but will make the implementation more confusing.

PR #1322, opened 2025-11-11
Implement basic rendering flow

### LINKED ISSUE #1271: Bevy Lifecycle / simple rendering
The goal here will just be to make sure that we can have Bevy present to the surface managed by Processing. 
Create the FFI code for driving the Bevy event loop, the idea here will but that each logical "frame" we'll drive the Bevy event loop forward one or more ticks (i.e. run the Main schedule) as well as figure out how to feed and map any user input events into the ECS.  This probably isn't strictly necessary at first as we won't be writing too many of our own user-responsive systems in Rust, but could be useful for things like render picking in the future. I think this will be the first step in which we stub out a class for WebGPU that extends `PGraphics`

tychedelia:
# Rendering

Closes #1271 

Processing has a rendering lifecycle for each `PGraphics` instance:
- `beginDraw`: called at beginning of frame or manually by the user to initialize the drawing state for a given surface. For us, this doesn't do much at the moment.
- `flush`: this lifecyle hook was added for OpenGL in order to flush accumulated state to the gpu. We'll use this similarly to render the currently accumulated draw state. TBD if the appropriate flushes are added everywhere we need them relative to what OpenGL does.
- `endDraw`: called at the end of frame or manually by the user to write to the surface/render target.

## Bevy's rendering loop

Bevy uses pipelined rendering. Specifically, there is an ECS "main" world and "render" world that run on separate timelines. We've currently disabled ECS multi-threading to ensure comparability, which means that the render world schedule runs after the main world serially, but conceptually they should be understood as independent. This isn't relevant for reviewing this PR but is an important implementation detail to know in general.

A `Camera` in Bevy represents a coarse grained unit of rendering work tied to an output render target. In other words, `Camera` == `PGraphics` and `RenderTarget` == `PSurface` for our concerns. Bevy *really* want to batch, both cameras and render state internal to cameras (i.e. opaque render items). On the contrary, processing really wants to force immediate mode.

In Bevy, every camera renders to a `ViewTarget`, which is an intermediate texture that accumulates rendering state while walking the render graph. At the end of the render graph, that texture is blitted to the `RenderTarget` which will be presented at the end of the frame.

This is important for understanding how clear state and `ClearColorConfig` works.... The `clear_color` field on `Camera` controls the clear for the *internal* rendering texture. Right now we are setting this with background color, but it should basically always be load in order to preserve the sketch across flushes. The `output_mode` field contains `CameraOutputMode`, which controls whether that internal texture is written to the render target at the end of the processing the camera. For us, we only want to set `CameraOutputMode::Write` on `endDraw`, which means we set `Skip` by default to preserve immediate mode semantics.

## Bridging immediate mode and batching

This PR faithfully implements the processing lifecycle, which requires a bit of juggling to ensure that we are only ever processing one `Camera` at a time. This will be important for preserving backwards compatibility with #1320, although we may be able to do additional optimization in the future.

Every time the user calls a processing API that needs to update the draw state, we record a `DrawCommand` that is stored in a buffer on the surface (typically the Window entity, this PR does not yet support off-screen rendering).

At the start of every `App` `update`, we ensure that all cameras are disabled by setting `active = false`, this ensures that regardless of the configured `output_mode`, that camera will not be processed for rendering. In Bevy terms, this guarantees the camera is not extracted to the render world.

When `flush` is called, we insert a market component `Flushing` on the surface that we intend to flush. First, we clear any previous meshes associated with this surface.  Then, we ensure the camera is set to active and thus we process only that camera.

We then drain the command buffer for that camera and render into a new set of meshes. We are using the `lyon` library to handle tessellation. See that libraries docs to understand how it works.

We currently implement some very simple batching logic meant to preserve the painter's algo used by Processing. Basically, we'll continue rendering into what I affectionately call a "mega mesh" that contains all the primitives (i.e. vertex data) for items that share the same material state. In this way, vertex order preserves the imperative draw order. We also apply a small z offset to ensure that when breaking up batches, we still draw in the correct order.

Right now, we don't really support much material configuration besides setting the alpha mode. This may change in the future when we enable users to write their own materials.

### Entity hierarchy

We're currently using the following entity hierarchy to help keep track of state per surface:

- `(Window, CommandBuffer)`: This is our "root" entity and the entity id that the Java side stores.
    - `Camera`: The camera configured for this surface which points to the surface as render target. This is what "sees" everything that we render for a given `PGraphics` as controlled by the `RenderLayers` we insert on both camera and mesh.
    - `TransientMesh` these (potentially multiple children) are meshes that we render into for this `PGraphics`. Right now we clear them at the beginning of each draw cycle.

# Testing

You can run the new `WebGPU.java` example to easily test.

# Minor changes

- Formatted all our rust files. Oops! We should probably add a CI hook for this. I use the following nightly options: `imports_granularity` and `group_imports` to make the imports clean.
- Renamed window to surface in most places. Technically in the future a surface may also be an off screen texture, so this is more accurate.

### DIFF
--- a/core/examples/src/main/java/WebGPU.java
+++ b/core/examples/src/main/java/WebGPU.java
@@ -1,0 +1,58 @@
+import processing.core.PApplet;
+
+public class WebGPU extends PApplet {
+    public void settings() {
+        size(600, 400, WEBGPU);
+    }
+
+    public void draw() {
+        background(200);
+
+        noStroke();
+        fill(255, 0, 0);
+        rect(50, 50, 80, 80);
+
+        fill(0, 255, 0);
+        rect(150, 50, 80, 80);
+
+        fill(0, 0, 255);
+        rect(250, 50, 80, 80);
+
+        fill(255, 0, 0, 128);
+        rect(50, 150, 80, 80);
+
+        fill(0, 255, 0, 128);
+        rect(150, 150, 80, 80);
+
+        fill(0, 0, 255, 128);
+        rect(250, 150, 80, 80);
+
+        stroke(0);
+        strokeWeight(4);
+        fill(255, 200, 0);
+        rect(50, 250, 80, 80);
+
+        fill(200, 0, 255);
+        rect(150, 250, 80, 80);
+
+        noFill();
+        stroke(255, 0, 255);
+        strokeWeight(6);
+        rect(250, 250, 80, 80);
+
+        noStroke();
+        fill(255, 0, 0, 100);
+        rect(400, 100, 100, 100);
+
+        fill(0, 255, 0, 100);
+        rect(440, 140, 100, 100);
+
+        fill(0, 0, 255, 100);
+        rect(420, 180, 100, 100);
+    }
+
+    public static void main(String[] args) {
+        PApplet.disableAWT = true;
+        PApplet.main(WebGPU.class.getName());
+    }
+}
--- a/core/src/processing/webgpu/PGraphicsWebGPU.java
+++ b/core/src/processing/webgpu/PGraphicsWebGPU.java
@@ -4,7 +4,7 @@ import processing.core.PGraphics;
 import processing.core.PSurface;
 
 public class PGraphicsWebGPU extends PGraphics {
-    private long windowId = 0;
+    private long surfaceId = 0;
 
     @Override
     public PSurface createSurface() {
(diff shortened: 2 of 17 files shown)


core/examples/src/main/java/WebGPU.java
```diff
@@ -0,0 +1,58 @@
...
+
+        fill(0, 255, 0, 100);
+        rect(440, 140, 100, 100);
+
+        fill(0, 0, 255, 100);
+        rect(420, 180, 100, 100);
+    }
+
+    public static void main(String[] args) {
+        PApplet.disableAWT = true;
```
catilac, line 55: Ah great, so for now this is how people could set up their sketch

catilac:
I'm going to branch off of this to add linux raw window support. I could also just merge this into the feature branch. let me know @tychedelia

catilac (approved):
Yay this works, i'll follow up with a wayland PR

PR #1333, opened 2025-11-19
Add .kotlin to gitignore

Stefterv:
Add .kotlin to gitignore as these are build files and should not be commited


catilac (approved):
lgtm!

PR #1334, opened 2025-11-19
Decoupling `Base.getSketchbookFolder()`

Stefterv:
This PR aims to decouple `Base.getSketchbookFolder()` from all its dependencies 

Previously the following was necessary to set up to be able to read the sketchbook folder:
```kotlin
        val resources = File("../resources-bundled/common")
        if (!resources.exists()) {
            throw IllegalStateException("Resources folder not found at ${resources.absolutePath}")
        }
        System.setProperty("processing.resources.folder", resources.absolutePath)
        System.setProperty("java.awt.headless", "true")
        Base.setCommandLine()

        Preferences.init()
        Base.locateSketchbookFolder()
```

With this PR, one can freely read `Base.getSketchbookFolder()` and all of the required initialisation will happen automatically. e.g.

```kotlin
val folder = Base.getSketchbookFolder()
```

It will also clean up the double reference to Processing's language files by just including them in the resources system.
I would prefer to use java's resources going forward and try to move away from the bundled folder.

In general there would need to be a push towards cleaning up the side-effects of a lot of these files but that is very much out of scope for now and this way I can continue building on top of the existing infrastructure.


app/test/processing/app/SketchbookTest.kt
```diff
@@ -0,0 +1,25 @@
+package processing.app
+
+import kotlin.io.path.createTempDirectory
+import kotlin.test.Test
+import kotlin.test.assertEquals
+
+class SketchbookTest {
```
catilac, line 7: a test a test!

catilac (approved):
This must have been satisfying

PR #1335, opened 2025-11-19
Rewrite of the `Base.getSettingsFolder()` and `Platform.getSettingsFolder()`

Stefterv:
Rewrote both functions so they fit into a single file, negating the need for hopping around when looking into what this functionality does. Now all the errors that can occur can be quickly visible and it no longer generates random awt windows through the `Messages` class

TODO:
- [x] Write tests
- [x] Test on Linux
- [x] Test on Linux with `sudo`
- [ ] Test on Linux with `XDG_CONFIG_HOME`
- [x] Test on Windows
- [x] Text on macOS
- [x] Figure out a steady way to support portable versions
- [ ] Test Portable preferences
  - [x] macOS
  - [ ] Windows
  - [ ] Linux

### DIFF
--- a/app/src/processing/app/Base.java
+++ b/app/src/processing/app/Base.java
@@ -2216,6 +2216,10 @@ public class Base {
    * @deprecated use processing.utils.Settings.getFolder() instead, this method will invoke AWT
    */
   static public File getSettingsFolder() {
+      var override = getSettingsOverride();
+      if (override != null) {
+          return override;
+      }
       try {
           return processing.utils.Settings.getFolder();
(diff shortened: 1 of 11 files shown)


Stefterv:
One thing I just realised, that maybe it is a good idea to migrate the settings folders on each platform also to a more modern place (maybe Java can give us a location)

catilac:
lgtm! What is a good way for me to test this?

Stefterv:
It shouldn't really change anything, so if you check the Preferences screen and compare between the local build and 4.4.10 they should be the same!

catilac (approved):
it works!

Issue #1305, opened 2025-10-24
Wayland support for simple rendering


catilac:
The rust/ffi/bevy work needs to work on wayland

catilac:
Branching off #1322 to implement this

PR #1337, opened 2025-11-20
wayland support

catilac:
Addresses #1305 

I still need to add support for X11. But can do that in a later PR.

### DIFF
--- a/core/src/processing/webgpu/PSurfaceGLFW.java
+++ b/core/src/processing/webgpu/PSurfaceGLFW.java
@@ -87,11 +87,12 @@ public class PSurfaceGLFW implements PSurface {
             PWebGPU.init();
 
             long windowHandle = getWindowHandle();
+            long displayHandle = getDisplayHandle();
             int width = sketch.sketchWidth();
             int height = sketch.sketchHeight();
             float scaleFactor = sketch.sketchPixelDensity();
 
-            webgpu.initWebGPUSurface(windowHandle, width, height, scaleFactor);
+            webgpu.initWebGPUSurface(windowHandle, displayHandle, width, height, scaleFactor);
         }
     }
 
--- a/core/src/processing/webgpu/PSurfaceGLFW.java
+++ b/core/src/processing/webgpu/PSurfaceGLFW.java
@@ -1,11 +1,6 @@
 package processing.webgpu;
 
-import org.lwjgl.glfw.GLFW;
-import org.lwjgl.glfw.GLFWErrorCallback;
-import org.lwjgl.glfw.GLFWFramebufferSizeCallback;
-import org.lwjgl.glfw.GLFWNativeCocoa;
-import org.lwjgl.glfw.GLFWNativeWin32;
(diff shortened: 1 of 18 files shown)


core/src/processing/webgpu/PSurfaceGLFW.java
```diff
@@ -123,6 +122,24 @@ public long getWindowHandle() {
...
+        }
+    }
+
+    public long getDisplayHandle() {
+        if (Platform.get() == Platform.MACOSX) {
+            // TODO: Currently unsupported
+            return 0;
+        } else if (Platform.get() == Platform.WINDOWS) {
+            // TODO: Currently unsupported
+            return 0;
```
catilac, line 139: wasn't sure what to do here since we don't use display handle on MacOS and windows yet.

core/src/processing/webgpu/PSurfaceGLFW.java
```diff
@@ -123,6 +122,24 @@ public long getWindowHandle() {
...
+        } else if (Platform.get() == Platform.LINUX) {
+            // TODO: need to check if x11 or wayland
+            return GLFWNativeWayland.glfwGetWaylandDisplay();
```
catilac, line 142: i think i'll add x11 to this if we can test in the short term. otherwise it could go in another PR. 
but happy to just do it here as well. i think i just have to check for a null pointer for Wayland or X11.

libProcessing/renderer/src/lib.rs
```diff
@@ -132,7 +133,7 @@ pub fn create_surface(
             }
         };
 
-        let window = AppKitWindowHandle::new(std::ptr::NonNull::new(ns_view_ptr).unwrap());
+        let window = AppKitWindowHandle::new(NonNull::new(ns_view_ptr).unwrap());
```
catilac, line 136: @tychedelia from a rust stylistic point of view, do we like removing namespacing? or better to keep it in the context of a larger program?

libProcessing/renderer/src/lib.rs
```diff
@@ -178,8 +179,20 @@ pub fn create_surface(
...
-    let (raw_window_handle, raw_display_handle) =
-        { todo!("implement linux raw window handle conversion") };
+    let (raw_window_handle, raw_display_handle) = {
+        use raw_window_handle::{WaylandDisplayHandle, WaylandWindowHandle};
+
+        let window_handle_ptr = unsafe { NonNull::new_unchecked(window_handle as *mut c_void) };
+        let window = WaylandWindowHandle::new(window_handle_ptr);
+
+        let display_handle_ptr = unsafe { NonNull::new_unchecked(display_handle as *mut c_void) };
+        let display = WaylandDisplayHandle::new(display_handle_ptr);
```
catilac, line 189: I should add a TODO here for X11 as well. But happy I figured out how to do this!~

catilac:
Closing this since the work has been moved to:
https://github.com/processing/processing4/pull/1343
https://github.com/processing/libprocessing/pull/3

PR #1340, opened 2025-11-21
libprocessing submodule

catilac:
`libprocessing` (all lower case) is now residing here: https://github.com/processing/libprocessing

and we have a new submodule!

PR #1343, opened 2025-11-24
Wayland Support

catilac:
Wayland support. This is based off of #1337 (which will now be closed since moving libprocessing). 

Also we need to merge in the submodule work first: 
https://github.com/processing/processing4/pull/1340


tychedelia:
Gradle build needs to be fixed separately.

PR #1345, opened 2025-11-25
Also look into the modules folder for native jars

### LINKED ISSUE #1327: JavaFX Requires 'code' Folder to Run

Build process

### Feature to enhance

Pre-installation of javafx.xxxx.jar files in another folder.

### Enhancement details

Pull request #856 added all seven of the javafx modules to the Processing editor.  However, some users particularly on Linux and Windows were unable to run JavaFX demos without manually adding a 'code' folder to the sketch and copy/pasting the jar files or drag 'n dropping the javafx.jar files onto an open sketch to create the 'code' folder automatically.  The javafx.xxxxx.jar files were present, but buried deeper inside operating system folders and inside a nested 'modules' folder.  Unfortunately the runtime did not know how to access these files.  The issue has been discussed in detail here: https://discourse.processing.org/t/processing-and-javafx/45692/13

**Related Finding:**
I've never had a problem running JavaFX demos on my MacOS system.  I now realize that this is because I have always had the javafx.xxxx.jar files in my _Processing/libraries/javafx/library_ folder.  I tried temporarily removing all of these and I was no longer able to run JavaFX files (missing library error message).  I then reinstalled the javafx.xxxx.jar files to this folder which 'fixed' the problem and brought back the ability to run Java/FX source code.

**Solution:**
A simple solution is to copy/paste all of the 'javafx.xxxx.jar' files plus 'javafx.jar'  into the _Processing/libraries/javafx/library_ folder where they are accessible to the runtime.  This has been tested on all three platforms and appears to be a valid solution to the problem.  To spare the user from copy/pasting the javafx.jar files perhaps they could be installed in the _Processing/libraries/javafx/library_ folder before distribution or when the library is installed from the Processing menubar (Sketch/Import Library.../JavaFX).  It is unclear to me when the 'javafx' folder is added to Processing/libraries.

**Confirmation:**
After installing the javafx modules as described above you should be able to run this simple demo on your system without using a 'code' folder:

**Test javafx source code:**

```Processing
import javafx.scene.canvas.Canvas;
import javafx.scene.layout.StackPane;

Canvas canvas;
StackPane root;

void setup(){
 size(350,200,FX2D);  
 surface.setTitle("JavaFX in Default Wnd");
 canvas = (Canvas)surface.getNative();
 root = (StackPane)canvas.getParent();
}

void draw(){
  fill(0,255,0);
  circle(width/2,100,100);
}

```

**Processing/libraries/javafx/library folder should look like this on Windows 11:**

Image

Yes, I’d like to help with this

### LINKED ISSUE #1344: JavaFX modules are not added to the classpath.

Build process

### Processing version

4.4.10

### Operating system

Windows 10

### Bug description

JavaFX modules are not included in the classpath by default.
These modules are required to run JavaFX programs.

The expectation is that JavaFX programs run without issues

### Steps to reproduce this

1. Install the JavaFX library.

3. Run the sketch snippet; the output will show that the JavaFX modules are not in the classpath.

### snippet

``` processing
import processing.javafx.*;

String classPaths = System.getProperty("java.class.path");

println(classPaths);
println();

String [] path = classPaths.split(";"); 
printArray(path);
```

### Additional context

A search for JavaFX related issues in these repositories will provide some history and context:

https://github.com/processing/processing4
https://github.com/processing/processing4-javafx

Related discussion on Processing discourse:

https://discourse.processing.org/t/processing-and-javafx/45692/16

I had some success modifying the Processing source code and adding the modules to the classpath during a collaborative session with Google Gemini.
I can share the details on request.

No, I’m just reporting the issue

Stefterv:
This change will make sure that Processing will scan the `$library/library/$native/modules` folder within a library folder, fixing the issues with JavaFX not having access to its modules

Closes #1344 
Closes #1327

### DIFF
--- a/app/src/processing/app/Library.java
+++ b/app/src/processing/app/Library.java
@@ -1,13 +1,16 @@
 package processing.app;
 
-import java.io.*;
-import java.util.*;
-import java.util.zip.ZipFile;
-
-import processing.app.contrib.*;
-import processing.core.*;
+import processing.app.contrib.ContributionType;
+import processing.app.contrib.LocalContribution;
+import processing.core.PApplet;
 import processing.data.StringDict;
 import processing.data.StringList;
+
+import java.io.File;
+import java.io.FilenameFilter;
+import java.io.IOException;
+import java.util.*;
+import java.util.zip.ZipFile;
 
 
 public class Library extends LocalContribution {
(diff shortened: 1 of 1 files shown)


app/src/processing/app/Library.java
```diff
@@ -398,27 +401,24 @@ public String getJarPath() {
...
-        cp.append(new File(libraryFolder, jar).getAbsolutePath());
-      }
-    }
+      addJars(libraryFolder, cp);
     File nativeLibraryFolder = new File(nativeLibraryPath);
     if (!libraryFolder.equals(nativeLibraryFolder)) {
-      jarHeads = new File(nativeLibraryPath).list(jarFilter);
-      if (jarHeads != null) {
-        for (String jar : jarHeads) {
-          cp.append(File.pathSeparatorChar);
```
Stefterv, line 414: Since I am in here, I also cleaned up some duplicated code

catilac (approved):
Yay! Thanks so much for your effort on this

PR #1346, opened 2025-11-25
Gradle fixes from libprocessing refactor

tychedelia:
Title. Tried to make things a bit more defensive using globs so we don't have to modify the build constantly for new crates we add.


catilac (approved):
yay thank you!

PR #1347, opened 2025-11-27
Gradle Plugin for creating Processing Libraries

Stefterv:
This PR will add a Gradle Plugin for defining Processing/Java libraries, it reduces the necessary setup of creating a new Processing library to following in a `build.gradle.kts` file. 

      plugins{
          id("org.processing.library")
      }
      
      processing {
          library {
              version = 1
              prettyVersion = "1.0.0"
      
              authors = mapOf(
                  "The Processing Foundation" to "https://processing.org"
              )
              url = "https://processing.org/"
              categories = listOf("file", "exporter", "dxf")
      
              sentence = "DXF export library for Processing"
              paragraph =
                  "This library allows you to export your Processing drawings as DXF files, which can be opened in CAD applications."
      
          }
      }

This will add the functionality of bundling a Processing library as a `.zip` and `.pdex` as well as functionality to install the new library directly into Processing to test.

I think we should create similar plugins for modes and tools!

### TODO
*After merging*

- [x] Add Gradle Publishing logic
- [ ] Migrate integrated libraries to new plugin
- [ ] Find and test a way to connect the debugger connected to Gradle towards running sketches in Processing for easy debugging when writing libraries. 
- [ ] After publishing, update the gradle library template. 

@tychedelia I've put the plugin in `gradle/plugins` rather than `buildSrc` as I'd like to avoid growing the top level folders in the repo, curious what you think about that or if I should rather stick to convention? So far functionality wise it seems the same.

### DIFF
--- a/gradle/plugins/library/build.gradle.kts
+++ b/gradle/plugins/library/build.gradle.kts
@@ -1,0 +1,28 @@
+plugins {
+    `java-gradle-plugin`
+    kotlin("jvm") version "2.2.20"
+}
+
+gradlePlugin {
+    plugins {
+        create("processing.library") {
+            id = "org.processing.library"
+            implementationClass = "ProcessingLibraryPlugin"
+        }
+    }
+}
+
+repositories {
+    mavenCentral()
+}
+
+dependencies {
+    testImplementation(kotlin("test"))
+}
+
+tasks.test {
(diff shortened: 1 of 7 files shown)


catilac (approved):
approved merge when ready @Stefterv

PR #1356, opened 2025-12-12
Reload legacy Preferences after saving

Stefterv:
Fixes this bug reported on the discord:

https://github.com/user-attachments/assets/ae8eb77d-aef9-4015-904f-fbcc7215d223

Calls Preferences.init() after saving preferences to ensure legacy Preferences are reloaded and kept in sync with the latest changes. Otherwise the new preferences state from compose would be overridden from the legacy system


catilac (approved):
I'm approving this, but expecting it to be tested by you and community. I'll run your next build!

PR #1360, opened 2025-12-15
Load supported locales from locales.txt file

Stefterv:
Refactors Language.listSupported() to read supported language codes from a new locales.txt resource file instead of scanning available locales.

Fixes a bug where the language files were not loaded correctly.


catilac (approved):
all the language show up and i can switch to them!

PR #1367, opened 2025-12-23
Fix tabbing issue with textfields

Stefterv:
Context: 
https://discourse.processing.org/t/processing-4-5-beta-is-out/47637/6

Changes textfields to single lines so they don't capture the tab-indexing


catilac (approved):
merged

PR #1381, opened 2026-01-15
Added `rebuildToolList()` to fix nullpointer in contributions check

Stefterv:
I just discovered that in 4.5.0 the contributions update didn't actually work, as the order of the logic changed in https://github.com/processing/processing4/pull/1299 and the tools list did not get build before it was being read.
In general I think we should really start rebuilding the startup logic of the PDE, as currently it is really hard to understand the interdependencies between different sections, also these sections do not work standalone at all.


catilac (approved):
I hear you. @Stefterv would you be able to create an issue describing what the startup work entails?

PR #1385, opened 2026-01-15
Reset defaults by setting value rather than remove

Stefterv:
Initially reset to defaults was set up by removing the key from the preferences but that didn't work well upstream, changing the value to default instead.


catilac (approved):
lgtm!

PR #1386, opened 2026-01-15
Create `PreferencesEvents` to glue into base and other deps

Stefterv:
To avoid creating dependencies within Preferences.kt, create an event structure that will listen to changes triggered by saving the Preferences.


catilac (approved):
nice i love this design

PR #1392, opened 2026-01-28
Updates from libprocessing.

tychedelia:


catilac (approved):
Looks amazing, thank you so much. The code is also incredibly clean and readable.
Also very interesting to see the almost 1-1 mapping in how the bindings and methodologies emerge from libprocessing.

Issue #1388, opened 2026-01-26
Preferences not validating inputs


catilac:
### Most appropriate sub-area of Processing 4?

PDE

### Processing version

4.5.x

### Operating system

all of them

### Bug description

When going into the advanced preferences section, if you enter invalid inputs, the application will crash.
In order to fix this, you have to delete 

Here is an example:
editor.tabs.size=2/

I accidentally typed in a "/" and the program crashed. But deleting that entry from `preferences.txt` fixed it.

### Steps to reproduce this

1. Go to advanced settings

2. Go to a text field input (like tab size) and enter non-numeric input

3. This will create a parse error and crash the application

### snippet

```processing

// Paste your code below :)

void setup() {

}

void draw() {

}
```

### Would you like to work on the issue?

Yes, I’d like to help with this

avinxshKD:
hey @catilac  I'd like to work on this, can you pls assign it to me? thanks

SableRaf:
Hi @avinxshKD. Sorry, but this was a high-priority bug and we wanted to fix it quickly, so @catilac took care of it.

If you’re still looking for issues to get involved with, feel free to [get in touch with us on Discord](discord.processing.org).

catilac:
@avinxshKD yes i'm so sorry about that, please reach out :) there is lots of fun stuff to work on, including extensions of this work that i just merged in

avinxshKD:
> Hi [@avinxshKD](https://github.com/avinxshKD). Sorry, but this was a high-priority bug and we wanted to fix it quickly, so [@catilac](https://github.com/catilac) took care of it.
> 
> If you’re still looking for issues to get involved with, feel free to [get in touch with us on Discord](discord.processing.org).

Thanks for the update @SableRaf,  totally understandable since it was high priority. 
I’m really keen to contribute, I’ll explore other open issues and join Discord. Thanks

avinxshKD:
@catilac No worries at all @catilac totally get it 🙂
I’d love to help with the extensions or anything related that you just merged. I’ll take a look and also join Discord. Looking forward to contributing more thanks again for alll the support yall<3

PR #1393, opened 2026-01-29
Read preferences txt, if fails, read defaults, else really break

catilac:
This is a bandaid fix for the tab-spacing Advanced error that was reported.
I'm carving out a set of issues for a larger fix. 

@SableRaf can you please give this a little test?

partial fix of https://github.com/processing/processing4/issues/1388

### DIFF
--- a/app/src/processing/app/Preferences.java
+++ b/app/src/processing/app/Preferences.java
@@ -336,7 +336,15 @@ public class Preferences {
 
 
   static public int getInteger(String attribute /*, int defaultValue*/) {
-    return Integer.parseInt(get(attribute));
+    try {
+      return Integer.parseInt(get(attribute));
+    } catch (NumberFormatException err) {
+      try {
+        return Integer.parseInt(getDefault(attribute));
+      } catch (NumberFormatException err2) {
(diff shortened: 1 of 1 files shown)


SableRaf:
In a quick test it seems to work 👍 Thanks @catilac 

Here's what I did:

**Replicating**

1. Opened Processing 4.5.1
2. Opened the preferences
3. Searched for "tabs"
4. Set `editor.tabs.size` to `4/` (causes error in the console)
5.  Closed and re-opened. Got the error from https://github.com/processing/processing4/issues/1388

**Testing the PR**

1. Built from PR branch (runs successfullly)
2. Opened the preferences
3. Searched for "tabs"
4. Set `editor.tabs.size` to `4` (removed the trailing `/`)
5. Closed Processing
6. Re-opened 4.5.1. (starts successfully)

SableRaf (approved):

SableRaf:
@catilac What's your take on adding comments in the code to explain why a workaround like this one was needed?

catilac:
@SableRaf i think generally that's a great idea! For this particular patch it's more like coding with a so-called best practice so it's not something too out of the ordinary. Thanks so much for reviewing will merge!

PR #1234, opened 2025-09-08
Internal Gradle Runner

Stefterv:
This PR adds the internal gradle runner to `app` to support running full Gradle projects within the PDE

### Features
- Adds support for running Gradle Projects to the PDE
- Adds a new `processing sketch run` CLI command for compiling Processing sketches from the command line
- Adds a new way of bundling sketches for the current OS 
- Adds a more modern way of listening for exceptions within sketches. Tries to give more info on what went wrong and provides a baseline to build a new FES on top of.
- Adds a system for defining GradlePlugins (which could be extended in the future to wider Processing Plugins)
- Adds a new per sketch settings panel to manage sketch based preferences
- Adds an experimental HotReloading feature

### Current issues
- [x] Windows: does not like the way the System.err and System.out are routed
- [x] Windows: some issue with the native binaries causing running to crash
- [x] macOS: Exported application (Could not find agent library jdwp on the library path)
- [x] macOS: Unsigned jar in distributable (race condition?)
- [x] Windows: Library class not found errors
- [x] Windows: Example folders are readonly causing gradle to crash (using `--build-file` argument for gradle does not work)
- [x] Windows: Failed Gradle Task does not report errors
- [x] Windows: Library examples seem to not be marked readonly
- [x] Windows: `Scan Libraries` task fails on contributed examples
- [ ] Windows: Provider.forUseAtConfigurationTime warning
- [ ] Windows: OpenJDK stays active after killing the PDE
  - [ ] Windows: This even prevents a library from being uninstalled because the file is still in use
- [x] Rename windres doesnt run
- [x] Linux: P2D and P3D renderer broken (Incorrect JOGL dependencies)
- [x] Snap: Not sure what is wrong yet
  - [x] Home Directory not accessible for Gradle distribution download

### TODO
- [ ] Add support for the integrated libraries (simply publish to the internal repository)
- [ ] Determine if we want to keep the hot reload & android plugins

### How to test
- Switch to the PR Branch
- Run Processing
- Run a sketch, observe the note about the legacy build system in the console
- Go to Settings -> Sketches -> Enable `Use Modern Build System`
- Save changes
- Run the sketch again, observe the note about the modern build system in the console
  - Ideally everything says the same in terms of functionality

PR #1398, opened 2026-01-30
Only set Newt window icons on Windows

Stefterv:
Wrap the NewtFactory.setWindowIcons(res) call in a platform check so icons are only set when PApplet.platform == PConstants.WINDOWS. This avoids invoking the icon-setting behavior on non-Windows platforms. Also reorganize and consolidate import statements (add missing AWT/Swing, JOGL and Processing imports) to fix compilation and include required classes.

This solves the issue where I got a JOGL natives error on macOS (and probably Linux as well)

Also solves an issue for #1234

### DIFF
--- a/core/src/processing/opengl/PSurfaceJOGL.java
+++ b/core/src/processing/opengl/PSurfaceJOGL.java
@@ -24,13 +24,31 @@
 
 package processing.opengl;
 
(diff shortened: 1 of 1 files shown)


catilac (approved):
lgtm!

PR #1401, opened 2026-02-02
Add PR artifact uploads and comments

Stefterv:
Update the build workflow to handle pull request runs: make the main artifact upload conditional (skip for pull_request), add a separate upload step for PR artifacts with longer retention, and post a sticky PR comment containing a download link. Also adjust workflow permissions to allow pull-request write access and include updated artifact paths.

See bot output here https://github.com/Stefterv/processing4/pull/11#issuecomment-3847385548

PR #1403, opened 2026-02-03
Publish `org.processing.core.dxf` to Maven Central (Publish libraries to maven central)

Stefterv:
Context:
https://discord.com/channels/1076634729618624534/1229526727320146042/1468237872074068141

- [x] DXF

Test is currently running on my repo results should show up here: https://central.sonatype.com/search?q=g%3Anl.steftervelde.java.libraries

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -3,6 +3,7 @@ plugins {
 
     alias(libs.plugins.compose.compiler) apply false
     alias(libs.plugins.jetbrainsCompose) apply false
+    alias(libs.plugins.mavenPublish) apply false
 
     alias(libs.plugins.versions)
 }
--- a/java/libraries/dxf/build.gradle.kts
+++ b/java/libraries/dxf/build.gradle.kts
@@ -36,7 +36,7 @@ dependencies{
 mavenPublishing {
     publishToMavenCentral(com.vanniktech.maven.publish.SonatypeHost.CENTRAL_PORTAL, automaticRelease = true)
     signAllPublications()
-    coordinates("$group.java.libraries", name, version.toString())
+    coordinates("$group.core", name, version.toString())
 
     pom {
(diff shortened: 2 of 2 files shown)


Stefterv:
@avinxshKD would you be interested in applying the same to some of the other Java Libraries?

avinxshKD:
@Stefterv Sure, I can take a look at those.

Breakdown looks like this:


- PDF: Good to go. It already has the Gradle setup, just needs the publishing block.

- Net / SVG: Need to be fully ported from Ant to Gradle first, but otherwise straightforward.

- Serial / IO: These are harder because of the native libs/JARs. We'll need to figure out how to package those correctly for Maven so we don't break cross-platform support.

One catch on this PR: [build.gradle.kts](vscode-file://vscode-app/c:/Users/avina/AppData/Local/Programs/Microsoft%20VS%20Code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) lists implementation("com.lowagie:itext:2.1.7"). pretty sure that's a copy-paste error from the PDF library? DXF shouldn't depend on iText.

Stefterv:
Great! Just make individual PR's for each package starting with the easiest. 

And thanks for catching the dependency!

catilac:
@Stefterv now that the action has been updated (#1401) , would it be a big lift to rebase this onto main? Or actually would a pulling in the changes from main suffice?

Stefterv:
Trying with a rebase on this and the PR before

avinxshKD:
@Stefterv @catilac 

 working on Serial and IO libraries now. PR coming soon. :)

avinxshKD:
Using the same coordinates as PDF/Net/SVG does that look right for consistency? @Stefterv

Stefterv:
Yes, `org.processing.core.serial` and `org.processing.core.io` please

PR #1409, opened 2026-02-05
use github.sha instead of ref_name

catilac:
I want to fix the broken builds having to do with the '/' character. An artifact of using `github.ref_name` on branches that aren't merged in yet. `github.sha` will give us a unique identifier

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -65,6 +65,6 @@ jobs:
       - name: Add artifact
         uses: actions/upload-artifact@v4
         with:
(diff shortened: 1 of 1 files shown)


Stefterv:
https://github.com/processing/processing4/pull/1401 fixes it in a similar manner and creates comments for people to download the preview builds, I read my description again and that was not clear, sorry!

catilac:
Closing this as it's already fixed in #1401 (This is the one i meant to close @Stefterv 😭 )

PR #1423, opened 2026-02-08
Processing Gradle Plugin - Version sync

Stefterv:
I was looking at the Gradle plugin code and realised I had missed something I had intended to implement. 

This PR adds a writeVersion Gradle task that writes project.version to build/resources/main/version.properties, so the plugin can access the plugin version at runtime. This way the version number of the plugin and the used version of processing.core will be the same.

### DIFF
--- a/java/gradle/build.gradle.kts
+++ b/java/gradle/build.gradle.kts
@@ -39,3 +39,15 @@ publishing{
         }
     }
 }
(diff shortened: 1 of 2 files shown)


catilac (approved):
great idea! so much consistency and sanity in our future

PR #1430, opened 2026-02-11
Update BUILD.md with cleaning build instructions

SableRaf:
Added troubleshooting step for build issues in IntelliJ using `./gradlew clean`


BUILD.md
Stefterv, line 184: It does already exist here...

catilac:
@Stefterv @SableRaf looks like this can be closed

Issue #1432, opened 2026-02-12
Resource Leaks in processing/app/Util.java


SalmaneKhalili:
## Debugging:

Hello, I was looking around the codebase for the sake of getting familiar with, where i noticed a couple problematic resource usage implementations that do not guarantee closure with either finally or encompassing with try ().
In multiple instances, if n exception occurs (disk full, corrupt ZIP, network timeout, etc.) file handles remain open. I believe on Windows this prevents subsequent deletion and on any OS it can exhaust the system’s file‑descriptor limit after repeated use.

## Examples:
```java
  static public void unzip(File zipFile, File dest) throws IOException {
    FileInputStream fis = new FileInputStream(zipFile);
    CheckedInputStream checksum = new CheckedInputStream(fis, new Adler32());
    ZipInputStream zis = new ZipInputStream(new BufferedInputStream(checksum));
    ZipEntry entry;
    while ((entry = zis.getNextEntry()) != null) {
      final String name = entry.getName();
      if (!name.startsWith(("__MACOSX"))) {
        File currentFile = new File(dest, name);
        if (entry.isDirectory()) {
          currentFile.mkdirs();
        } else {
          File parentDir = currentFile.getParentFile();
          // Sometimes the directory entries aren't already created
          if (!parentDir.exists()) {
            parentDir.mkdirs();
          }
          currentFile.createNewFile();
          unzipEntry(zis, currentFile);
        }
      }
    }
  }
```
ZipInputStream, CheckedInputStream, and FileInputStream are never closed. i think this may also contribute to update issues if there are any, as any crash may lock the temporary zip file, preventing its deletion.

## A simple solution would be:
```java
public static void unzip(File zipFile, File dest) throws IOException {
  try (ZipInputStream zis = new ZipInputStream( new BufferedInputStream( new CheckedInputStream( new FileInputStream(zipFile), new Adler32())))) {
    ZipEntry entry;
    while ((entry = zis.getNextEntry()) != null) {
      //rest of code
    }
  }
}
```
this way all three underlying streams are closed automatically, even on exception.
This pattern repeats itself in other methods in the same file, some of which are: loadBytesRaw(File), saveFile(String, File) , copyFile(File, File).

## Willing to work on this?
If i may, i'd be more than happy to be assigned to this issue as my first contribution to this codebase!

SalmaneKhalili:
If this sort of proposals are not welcomed, please let me know so that i can go ahead and close it. @SableRaf

[...]

PR #1436, opened 2026-02-15
Resource Leaks in processing/app/Util.java

SalmaneKhalili:
This PR add a simple implementation of what was suggested in #1432, at the moment it was only applied to unzip, i should be able to do the same for the rest of the methods mentioned in the issue, but id like to have feedback on the current way im going about it.
- Added the test to gradle path
- added the actual test.
- added the solution


app/test/processing/app/UtilTest.java
tychedelia, line 1: Probably better to not have a test than a platform specific test.

tychedelia (commented):
Looks good! We should just either gate or remove the test.

app/test/processing/app/UtilTest.java
SalmaneKhalili: Done!

catilac:
Thanks y'all! Running the tests. I'll go ahead an approve it

catilac (approved):

catilac:
@all-contributors please add @SalmaneKhalili for bug-finding, and bug-fixing

PR #1437, opened 2026-02-17
Fix modal dialogs freezing all IDE windows (#1375)

### LINKED ISSUE #1375: Stop modal message from freezing all windows

I'm not sure

### Feature to enhance

Stop modal messages from freezing all windows

#### Enhancement description
Whenever a modal IDE message appears, it should not freeze the other windows of the IDE.
Or if so, it should always remain the frontmost window of the IDE.

#### Benefits
This would prevent IDE windows from being frozen with no explanation (modals can get lost in background).

#### Challenges
Reproducing window-based modality is probably tricky. I don't know about locking the modal window at the front.

#### Additional context
I had a case where windows were all frozen, and I did not notice the modal message.
I thought that it was a bug.

No, I’m just suggesting the enhancement

deekshithaby:
Problem
When a modal message appears (error, warning, "Save first", etc.), it freezes ALL Processing IDE windows. Users can't interact with any other window, and the modal can get lost behind other windows with no way to recover.

Root Cause
- `Messages.kt`: All popup dialogs used `Frame()` as their parent (an orphan frame with no connection to any editor window), causing `APPLICATION_MODAL`behaviour that blocks every window.
- `Base.java`, `Sketch.java`, `ShimAWT.java`: `FileDialog` instances did not set modality type, defaulting to `APPLICATION_MODAL`.

Fix
-Messages.kt: Created a `showModalDialog()` helper that finds the currently active window via `KeyboardFocusManager` and creates `DOCUMENT_MODAL` dialogs. Replaced all 5 `JOptionPane.showMessageDialog(Frame(), ...)` calls with this helper.
- **FileDialogs**: Set `DOCUMENT_MODAL` on all `FileDialog` instances in Base.java, Sketch.java, and ShimAWT.java.

Result
- Modal messages (errors, warnings, confirmations) now only block their parent window — other IDE windows remain fully interactive.
- File dialogs respect `DOCUMENT_MODAL` on Linux/Windows. Native macOS file dialogs ignore Java modality settings (Apple limitation), but all JOptionPane-based messages are fixed across all platforms.

Testing
- Opened 3+ editor windows
- Triggered error messages in each window
- Confirmed other windows remained responsive while popups were showing

Fixes #1375

### DIFF
--- a/app/src/processing/app/Base.java
+++ b/app/src/processing/app/Base.java
@@ -1155,6 +1155,7 @@ public class Base {
       // use the front-most window frame for placing file dialog
       FileDialog openDialog =
         new FileDialog(activeEditor, prompt, FileDialog.LOAD);
+        openDialog.setModalityType(Dialog.ModalityType.DOCUMENT_MODAL);
 
       // Only show .pde files as eligible bachelors
       openDialog.setFilenameFilter((dir, name) -> {
--- a/app/src/processing/app/Messages.kt
+++ b/app/src/processing/app/Messages.kt
@@ -48,6 +48,19 @@ import javax.swing.UIManager
 
 class Messages {
     companion object {
+
+        /**
+         * Shows a modal dialog that only blocks its parent window,
+         * not all windows in the application.
+         */
+        private fun showModalDialog(message: Any, title: String, messageType: Int) {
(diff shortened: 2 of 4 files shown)


catilac:
Awesome, thanks for finding this and fixing. will review

PR #1441, opened 2026-02-19
Resolves: #[1417] Value Now Compares to missingFloat Variable

### LINKED ISSUE #1417: Check for missing value
**[[Reserved for Mark Sherman's Class]]**

https://github.com/processing/processing4/blob/main/core/src/processing/data/Table.java#L4822

This is specifically about the `TODO`. 
There is a `missingFloat` field in this class. We should be comparing against that.

ericahin:
✅ **Title:** clear and descriptive

✅ **Resolves:** Resolves #1417

✅ **Changes:** Changed the comparison on line 4822 that is checking if the float variable "value" is NaN to use the globally declared variable "missingFloat." This was labeled as a TODO. This maintains consistency throughout the document so is anything were to change, this comparison would stay consistent with the rest of the file.

✅ **Tests:** I created a function in TableTest.java called testGetMaxFloat() which checks all edge cases of Null, negative numbers, 0, positive numbers, etc. I also tested using the "./gradlew test" command, all tests were passed by this code

✅ **Checklist:** All tests pass and I have pulled/pushed the most recent changes/updates from upstream onto my branch

### DIFF
--- a/core/src/processing/data/Table.java
+++ b/core/src/processing/data/Table.java
@@ -4819,7 +4819,7 @@ public class Table {
     for (int row = 0; row < getRowCount(); row++) {
       for (int col = 0; col < getColumnCount(); col++) {
         float value = getFloat(row, col);
-        if (!Float.isNaN(value)) {  // TODO no, this should be comparing to the missing value
+        if (Float.compare(value, missingFloat) != 0) { //value now compares to missingFloat variable
           if (!found) {
             max = value;
             found = true;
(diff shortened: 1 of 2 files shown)


marksherman:
@elisefergusonn can you please review

elisefergusonn:
LGTM 👍

srnkan:
Looks good to me too!

catilac (approved):
wonderful! thank you so much @ericahin

PR #1444, opened 2026-02-21
Add Boundary

LauraL0322:
Null pointer checks have been added to all places where null might occur in basic.java file:
Added null pointer protection: draw() method, drawInstructionBar() method, drawButton() method, mousePressed() method, keyPressed() method. 
The code now has full null pointer protection, so it won't crash even if some objects are not initialized correctly.


SableRaf:
Hi @LauraL0322 and thanks for opening this PR. Could you please link to the related issue if there is one?

Issue #1410, opened 2026-02-05
Every time I try to switch from Java Mode to Android Mode


fsamanos-boop:
### Most appropriate sub-area of Processing 4?

I'm not sure

### Processing version

4.5.2

### Operating system

Linux mint 22.1 Xia 

### Bug description

I get the following error: "java.awt.IllegalComponentStateException: The frame is displayable".
​This happens on both my desktop and my laptop. It seems to be related to java.awt.Frame.setUndecorated being called while the frame is already visible.
​Steps to reproduce:
​Open Processing.
​Click on the "Java" mode button to switch to "Android".
​The error window pops up immediately.

### Steps to reproduce this

1.

2.

3.

### snippet

```processing

// Paste your code below :)

void setup() {

}

void draw() {

}
```

### Would you like to work on the issue?

No, I’m just reporting the issue

SalmaneKhalili:
## Debugging
Hello, looking at the documentation, it seems frames cannot be set to undecorated while being displayed.

<img width="1911" height="808" alt="Image" src="https://github.com/user-attachments/assets/406c72da-acd6-4b20-802c-190c0eab4054" />

## Solution
A simple solution is to encompass the setUndecorated(true) method in Editor.java with a isDisplayable() conditional.
```java
      if (Platform.isLinux()) {
        if (!isDisplayable()) {
          setUndecorated(true);
        }
          getRootPane().setWindowDecorationStyle(JRootPane.FRAME);
      }
```
## TRADEOFF
EDIT - It is no longer being decorated, this PR is ready for user testing! @catilac

fsamanos-boop:
Thank you very much, I will try it

PR #1447, opened 2026-02-23
Bug/androidcrash

SalmaneKhalili:
## Issue
#1410
## Solution
Conditional using isDisplayable().


SalmaneKhalili:
Still pending work, just realized that i tradedoff the crash with the window not being undecorated, so ill be looking more into this.

SalmaneKhalili:
> Still pending work, just realized that i tradedoff the crash with the window not being undecorated, so ill be looking more into this.

The fix apparently was as simple as moving the linux check to the beginning of the constructor. it is now back to being properly undecorated + not crashing.

catilac (approved):

PR #1459, opened 2026-03-03
Gradle 9.0.0

catilac:
For some reason github actions is now using Gradle 9.0.0. Unfortunately I missed what exactly changed in that regard.
Locally we were using an earlier version. In this PR I upgraded to Gradle 9.0.0 and have attempted to make it work.

NOTE: I did end up using a combination of Claude code and internet searches to figure this one out.


java/gradle/build.gradle.kts
```diff
@@ -51,11 +51,11 @@ tasks.withType<Test>().configureEach {
 }
 
 tasks.register("writeVersion") {
-    // make the version available to the plugin at runtime by writing it to a properties file in the resources directory
     doLast {
+        val version = project.version.toString().takeIf { it != "unspecified" } ?: return@doLast
```
Stefterv, line 55: This will mean that the tests will run w core 4.3.4 instead of the locally build one

marksherman:
Confirming this does run tests, build, etc successfully for me locally (MacOS)

Stefterv:
As mentioned before on discord, unfortunately that is because it skips the failing test rather than solve it

PR #1466, opened 2026-03-04
Pull Request - User Story: Save customisation settings

prajaktasen:
This modification adds persistent interface customization to the Processing IDE editor window. The editor now saves and restores user interface layout settings between sessions.

The following settings are stored using the Preferences system:

Window position (X, Y)

Window size (width, height)

Divider position between the editor and console

Maximized window state

When the editor window is moved, resized, maximized, or when the console divider is adjusted, the layout settings are automatically saved. When Processing is reopened, these saved settings are applied so the editor appears exactly as the user left it.

Additionally, a new menu option “Reset Interface Customization” was added to the File menu, allowing users to clear the saved layout settings and restore the default editor layout.

This change improves usability by maintaining a consistent workspace across sessions.

### DIFF
--- a/app/src/processing/app/ui/Editor.java
+++ b/app/src/processing/app/ui/Editor.java
@@ -65,6 +65,13 @@ import java.util.stream.Collectors;
  * Main editor panel for the Processing Development Environment.
  */
 public abstract class Editor extends JFrame implements RunnerListener {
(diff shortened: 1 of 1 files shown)


catilac:
What issue is this for? Please post your strategy on that issue, or create an issue before filing a PR

PR #1400, opened 2026-02-02
Permissions Fix

Stefterv:
Set dir permission to fix an issue for the nix packagemanager

PR #1473, opened 2026-03-05
Fix JDK dirPermissions placement in gradle build files

neobrain:
This is required for building against read-only JDK installations (notably used when building with Nix). #1400 already attempted to integrate my nixpkgs patch for this purpose, but it accidentally placed the dirPermissions line under a different task.


Stefterv (approved):

catilac (approved):

neobrain:
@catilac Thanks for merging! I'll look into updating Processing on nixpkgs once the next release is out, now that we can drop the downstream patches :)

catilac:
yay! so excited about nix support @neobrain :)

PR #1477, opened 2026-03-11
Lower bounds check on IntList get

### LINKED ISSUE #1415: Lower bounds checking on IntList `get` method
**[[Reserved for Mark Sherman's Class]]**

In https://github.com/processing/processing4/blob/main/core/src/processing/data/IntList.java#L166-L171

We are not doing a lower bounds check on `index`. We don't know what would happen with negative indices.

1. Please write a test to see what happens when there is a negative index,
2. Update the test to check if we do the "right-thing" in that case (probably an exception? please figure it out! (or discuss with me :) ))
3. Update the code to make the test pass!

smarasca5:
Title: Lower bounds check on IntList get

Resolves #1415

Changes: Added a lower bound within the if statement of the get method within IntList. There was no lower bounds check within this method before. Now it throws exception if index is negative.

Tests: tests within IntListTest

Checklist: Tests pass and branch is up to date

### DIFF
--- a/core/src/processing/data/IntList.java
+++ b/core/src/processing/data/IntList.java
@@ -6,6 +6,7 @@ import java.util.Arrays;
 import java.util.Iterator;
 import java.util.Random;
 
+import org.jetbrains.annotations.TestOnly;
 import processing.core.PApplet;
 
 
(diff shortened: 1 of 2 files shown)


marksherman:
We might want to squash the commits into a single commit thaat has a more descriptive name. That way there's one commit with all the small changes packed together, so looking at the history it will be clear what happened here.

marksherman:
@FranciscoJ26 could you please review these changes

FranciscoJ26:
LGTM

smarasca5:
newly updated! now has tests :) now ready for review

catilac (approved):
Thank you so much @smarasca5 @marksherman

Issue #1470, opened 2026-03-04
Add unit tests for JSONArray


satindra-r:
### Most relevant area for this enhancement?

Unit testing

### Feature to enhance

`JSONArray` class in core

#### Enhancement description
The `JSONArray` class does not have any tests even though it is used in `app` and has substantial logic

#### Additional context
Part of #880 

### Would you like to help implement this enhancement?

Yes, I’d like to help with this

NANDGOPALSHARMA-29:
Hi! I’d like to work on this issue.
I’m currently setting up the Processing4 codebase locally and exploring the JSONArray class. I’d like to contribute by adding unit tests for it.
Please let me know if this issue is available to work on.

SableRaf:
Hello @NANDGOPALSHARMA-29 

It looks like @satindra-r has already expressed interest in working on this issue. We usually follow a first-come first-served approach to issues to avoid duplicated work. 

Also, before work can start, maintainers typically need to confirm the issue and agree on an implementation.

Please consult our `CONTRIBUTING.md` for detailed guidelines.

---
 
_**A note on GSoC:** We receive many contributions around this time of year from GSoC applicants. Just a heads-up that our application process does **not** require you to submit a PR. Please check out [this thread on the forum](https://discourse.processing.org/t/updates-about-gsoc-2026-org-application-in-progress/47450) for more details on how to apply._

satindra-r:
Hi, is there anything specific regarding the implementation of the tests that I have to confirm before being assigned?

satindra-r:
I mainly want to test the constructor and the `writeInternal` functions as it contains JSON parsing logic which is fragile

catilac:
@satindra-r i've gone ahead and assigned you

PR #1484, opened 2026-03-16
Add parsing related unit tests for JSONArray

satindra-r:
This PR adds tests to the `JSONArray` class in core which has some fragile parsing logic.

Solves #1470

### DIFF
--- a/core/test/processing/data/JSONArrayTest.java
+++ b/core/test/processing/data/JSONArrayTest.java
@@ -1,0 +1,381 @@
+package processing.data;
+
+import org.junit.Test;
+
+import java.io.ByteArrayOutputStream;
+import java.io.PrintWriter;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertThrows;
+
+
+public class JSONArrayTest {
+  @Test
+  public void testDefaultConstructor() {
+    JSONArray jsonArray = new JSONArray();
+    assertEquals(0, jsonArray.size());
+  }
+
+  @Test
+  public void testEmptyTokenerConstructor() {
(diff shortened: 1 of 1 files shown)


core/test/processing/data/JSONArrayTest.java
```diff
@@ -0,0 +1,363 @@
...
+    assertEquals(1, jsonArray.get(0));
+    assertEquals(2, jsonArray.get(1));
+  }
+
+  @Test
+  public void testNullFormat() {
+    String json = "[\n" +
+            "null,\n" +
+            "null\n" +
+            "]";
```
catilac: why not just make this one string instead of multiple addition operations?

core/test/processing/data/JSONArrayTest.java
```diff
@@ -0,0 +1,363 @@
...
+    String output = jsonArray.format(0);
+    assertEquals(json, output);
+  }
+
+  @Test
+  public void testIntFormat() {
+    String json = "[\n" +
+            "1,\n" +
+            "2\n" +
+            "]";
```
catilac: same as above, this only makes things more difficult to read

catilac (changes requested):
lgtm, i've added a comment about some stylistic changes!

catilac:
@satindra-r thank you so much! i've left some feedback, and i'm triggering the tests

core/test/processing/data/JSONArrayTest.java
satindra-r: Oops, I forgot that raw strings were added in java 13. I've made all the json strings into raw strings

PR #1494, opened 2026-04-01
Added testing for Sort.java

ericahin:
**Tests/Changes:** Added tests for the Sort.java and the PDEWelcome.kt file since it doesn't currently have any. I tested all them and they all pass. I also falsified and tested again! I used Claude to generate ideas of what to test and some comment text, but all the actual code is self written.


marksherman:
LGTM

catilac (approved):
yes!

Issue #1490, opened 2026-03-27
Contributors overview image for the README.md


xinemata:
### Relevant sub-area for this feature?

Accessibility, Other (specify if possible)

#### Feature description
In order to increase Processing 4 contributor community's visibility, it'd be great to follow p5.js' footstep to create a "contributors overview image" that can represent all Processing 4 contributors. This image could be used in contexts like such as Processing's README.md or [Processing Foundation's people page](https://processingfoundation.org/people). See [p5.js' readme](https://github.com/processing/p5.js?tab=readme-ov-file#contributors) as a parallel example. 

![p5.js contributors overview image](https://raw.githubusercontent.com/processing/p5.js/refs/heads/main/contributors.png)

#### Benefits
Creating more visibility for Processing contributors.

#### Challenges
I'm not sure what the challenges may be.

#### Additional context
For p5.js, this feature was created through [#8237](https://github.com/processing/p5.js/issues/8237). 

### Would you like to help implement this feature?

No, I’m just suggesting the feature

catilac:
Oh great idea! I'm going to put this link here too: https://github.com/processing/processing4/blob/main/CONTRIBUTORS.md

susan-t:
@xinemata Hi! I'm new and would love to work on this issue! If it’s available, could I?

catilac:
@susan-t make a suggestion for how you would approach it and i'll assign!

Hemant210:
Hi! I’d love to work on this issue. I’ll start by exploring how p5.js implements the contributors overview image and propose a similar approach for Processing4.

susan-t:
I would have the same approach as @Hemant210. Would love to work on it with them if it works out?

Hemant210:
Hi @susan-t , great to see we're on the same page!
Maybe we can explore this separately and keep sharing our ideas and progress here?

catilac:
I'm going to assign this to @kjhollen, we had a discord conversation and i would like this be taken care of sooner than later.

Moreover, I want to be clear that no solution has actually been proposed.

catilac:
This is completed

PR #1497, opened 2026-04-08
add github action to implement #1490 contributor image

kjhollen:
Implements @skyash-dev's solution from the p5.js repository:

- add a github action to generate the image in contributors-png.yml
  - add utils/contributors-png.js to use node & canvas to make an image
  - use processing-bot to generate a pull request with the new updated image when the all contributors file is changed
  - action only applies to processing/processing4 repo's main branch
- add contributors.png file with current list of contributors (generated during testing)
- make contributors.png visible in the README in the contributors section


ksen0:
Definitely not a blocker: but potentially it's worth abstracting out both this and the p5 action into a separate action repo, like this one: https://github.com/processing/branch-pr-close-issue so the code is not duplicated in 2 places. @skyash-dev let me know if that would be of interest to you since you made the original action? I can make the repo and you could work on adding the action there and then updating the usages in processing4 and p5.js repos to use that.

skyash-dev:
@ksen0 sounds good to me, happy to take this on.

ksen0:
@skyash-dev great, thank you! I've made a repo here: https://github.com/processing/all-contributors-image/issues any of those are yours if you want them. I can only assign an issue after you comment on it. (You're welcome to suggest a new name for it as well)

catilac (approved):

PR #1503, opened 2026-04-13
P Font Tests: ps name getter and size getter, windows and mac

smarasca5:
P Font Tests: ps name getter and size getter, windows and mac

Changes: Added tests within PFontTest.java for ps name getter and size getter 

Tests pass and branch is up to date, commits have been squashed


catilac:
I already merged the other PFontTests you wrote! I think this PR may have been left by accident

PR #1506, opened 2026-04-19
Creating tests for WelcomeSurvey.kt

srnkan:
Changes/Tests: I added tests for the WelcomeSurvey.kt data file, since there were no preexisting tests for the file. Subsequently, a new test folder was created.  All tests pass and Claude was used for the general test outline.


SableRaf:
We are turning off this component in 4.5.3, and the text will likely be different if we ever turn it back on, so I don’t think adding tests for the current survey copy makes sense at this point. Thanks for taking a look at it though!

cc @kjhollen @catilac

catilac:
Thanks for speaking up on this @SableRaf !

Issue #1507, opened 2026-04-20
Mac release build fails with code signing issue


kjhollen:
### Most appropriate sub-area of Processing 4?

Build process

### Processing version

4.5.3

### Operating system

macOS

### Bug description

Release level builds fail on macOS at the notarization step (see [last release build run output](https://github.com/processing/processing4/actions/runs/22582033347/job/69365516412)) Testing locally, when you run:

```
./gradlew app:packageDistributionForCurrentOS
```

And then examine Processing.app with the codesign tool:

```
codesign --verify --deep --strict --verbose=2 <path-to-processing>
```

The output says "a sealed resource is missing or invalid" and that the resources for pde.icns, pdez.icns, and pdex.icns have been added since the app was signed.

### Steps to reproduce this

1. Build on mac with `./gradlew app:packageDistributionForCurrentOS`

2. Run `codesign --verify --deep --strict --verbose=2 <path-to-processing>` on the .app created in the build process

### snippet

```processing

// Paste your code below :)

void setup() {

}

void draw() {

}
```

### Additional context

With thanks to Claude and StackOverflow, I've learned that Jetpack Compose's default macOS app signing is not compatible with `fileAssociation()` so we either need to:

1. add a step to the process that re-signs the macOS app after the files have been associated, but before they are sent to the notary OR
2. separate how Windows, Linux, and Mac handle file associations so we can keep using the default Jetpack Compose app signing

I have a tentative solution that looks like option 2. I'll share a draft—I only have access to a Mac for development and can't test Windows & Linux on my own.

### Would you like to work on the issue?

Yes, I’d like to help with this

SableRaf:
Thanks @kjhollen for investigating this! 💙

Could you share the link to the StackOverflow discussion? I wasn’t able to find it.

Not sure if this is directly relevant, but sharing for reference: It looks like support for `fileAssociation()` was implemented in JetPack Compose (see [JetBrains/compose-multiplatform/issues/773](https://github.com/JetBrains/compose-multiplatform/issues/773) ) in 2024. That particular issue doesn’t mention code signing though.

kjhollen:
@SableRaf StackOverflow was more for helping me (partially) understand the Jetpack compose build system, but it was Claude/Copilot that suggested the fileAssociation issue. But, it seems strange that this wouldn't be listed as an issue, if true—probably a hallucination? I'll see if I can do some more digging today.

kjhollen:
[...]

PR #1510, opened 2026-04-27
fix for #1507 - mac signing issues for release builds

kjhollen:
fixes issue #1507 with Mac app notarization failure by signing the whole app again after fileAssocations are made. Adds an extra step called 'signApp' to the build process for `./gradlew -m packageDistributionForCurrentOS` which depends on `createDistributable` and `setExecutablePermissions`.

Previously, release builds were failing because the .icns files for .pde, .pdex, and .pdez were added to the .app file after signing, when `fileAssocation` was called in `nativeDistributions`.

Most of this code was generated with Copilot, then I edited it for legibility. I asked Copilot a lot of questions about the order of operations for the build system and where the fileAssociations occurred, and feel confident that this is the right place in the process to re-sign the app. I also prodded Copilot to see if there were any other simpler options that didn't involve signing things a second time (outside of when jpackage does this automatically), but wasn't able to produce any other successful solutions.

I have only tested this on Mac, but the file type associations work and the app notarization completed successfully (though it took nearly 30 minutes to run the notarization).

pinging @catilac for review!

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -186,7 +186,7 @@ tasks.register<Exec>("packageCustomDmg"){
     onlyIf { OperatingSystem.current().isMacOsX }
     group = "compose desktop"
 
-    dependsOn(distributable(), "installCreateDmg")
+    dependsOn("signApp", "installCreateDmg")
 
     val packageName = distributable().packageName.get()
     val dir = distributable().destinationDir.get()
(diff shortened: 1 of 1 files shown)


Stefterv (approved):

Stefterv:
It ain't pretty, but if it works, it works!

I created a release here to test if it works in the CI: https://github.com/Stefterv/processing4/releases/tag/processing-1448-4.5.18

And yes, notarisation always takes around 15-25 minutes

Looks like it works!

catilac (approved):
cool! looks good!

PR #1512, opened 2026-04-27
Compose Top Bar Menu

smarasca5:
kotlin top bar with pop up menus

Work in Progress

we wrote a topbar in compose, this is the first attempt at converting certain parts of the editor into Kotlin. The top bar is a compose instance that uses the existing JMenu logic to pop up menus after clicking on the top bar. No tests have been written so far


catilac:
@smarasca5 @Ebaron96 can you share a screen shot? is this still a WIP?

PR #1525, opened 2026-06-12
Autofill Comma Correction for Visual Studio Code Extension

John-Bobular:
In the current release version of the Processing Visual Studio Code extension, a comma is given for each argument that a method takes, rather than putting one comma. This means when auto filling a method call that has 2 arguments, 2 commas will be put inside the method parameters when only one would be needed to separate the two values.

### DIFF
--- a/java/src/processing/mode/java/lsp/PdeAdapter.java
+++ b/java/src/processing/mode/java/lsp/PdeAdapter.java
@@ -318,11 +318,15 @@ class PdeAdapter {
       for (char ch : chs) {
         if (ch == ',') {
           n += 1;
-          //insert += ",$" + n;
-          newInsert.append(",$").append(n);
+          /*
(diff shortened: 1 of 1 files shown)


SableRaf:
Hi @John-Bobular and thanks for reporting and proposing a fix! I'm trying to reproduce the issue but it's not totally clear to me what the actual and expected behaviors are. Would you mind sharing the steps needed and what I should look at? Before/after screenshots might help.

catilac:
@John-Bobular @SableRaf yes would love steps to reproduce what you're describing. It will help me to review. I have a sense, but I need to be able to re-create (and understand) what you're fixing. thank you!

catilac:
just pinging again @John-Bobular. I will close this PR  to keep things clean. if i end up closing it feel free to open up another PR or file an issue with details of what you're fixing so i can properly evaluate. thank you.

PR #1528, opened 2026-06-19
Fix REG_APP_DIR to support jpackage application launchers

tutimura-kg:
### Description
This PR fixes an issue where the file association fails when Processing is packaged and launched using `jpackage`.

Currently, `REG_APP_DIR` relies directly on `System.getProperty("user.dir")`. However, when launched via a `jpackage` generated launcher, `user.dir` points to the directory from which the application was invoked (e.g., the current terminal directory or shortcut location), rather than the actual installation directory of `processing.exe`. This causes incorrect registry paths to be registered in `checkAssociations()`.

Thus, launching via file association breaks the next launch via file association. And launching via shortcut fixes the file association. This has not surfaced as a major issue on Windows 11 because the File Manager can override registry-defined associations. However, it remains a hidden bug that this PR aims to fix.

### Changes

- Updated `REG_APP_DIR` to dynamically resolve the installation directory based on the location of the JAR file (via `WindowsPlatform.class`).
- Added a fallback to the original `user.dir` behavior in case `processing.exe` cannot be found relative to the JAR location.
- Kept the changes minimal to avoid any unexpected impact on the existing codebase.

This ensures that file associations (`.pde`, etc.) always register the correct absolute path of `processing.exe`, regardless of the working directory at launch.

### DIFF
--- a/app/src/processing/app/platform/WindowsPlatform.java
+++ b/app/src/processing/app/platform/WindowsPlatform.java
@@ -127,10 +127,10 @@ public class WindowsPlatform extends DefaultPlatform {
       File exeDir = jarDir.getParentFile();
 
       if (exeFileExists(exeDir)) return exeDir.getAbsolutePath(); // standard folder layout
-      if (exeFileExists(jarDir)) return jarDir.getAbsolutePath(); // fallback if exe is inside "app"
+      if (exeFileExists(jarDir)) return jarDir.getAbsolutePath(); // if exe and jar is in same folder
     } catch (java.net.URISyntaxException e) {
     }
-    return System.getProperty("user.dir");
+    return System.getProperty("user.dir"); // processing.exe not found
   }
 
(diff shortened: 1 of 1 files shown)


tutimura-kg:
The launcher executable created with jpackage continues to run as a process, so we can easily obtain the executable using `ProcessHandle.current().info().command().orElse(null)`. We can make use of it. In this case, during debugging (launched by `gradlew run`), it will be `...\java.exe`, so we will still need to substitute it with the current directory.

catilac:
Thanks @tutimura-kg. I am currently doing some fixes to the build system so we can commence releases again. Will review your work after that.

catilac:
@tutimura-kg okie! the build system is fixed. i will review soon :)

ntsuchimura:
Ok, I'll be waiting.
For reference, I used `regedit.exe` to check the path of the associated executable file. I needed to type [F5] to update.
processing-assoc

PR #1552, opened 2026-07-20
un-revert file permissions fix

catilac:


catilac:
@neobrain hi hi can you review?

neobrain (commented):
GitHub doesn't let me "approve" this, so my verbal approval will have to do: I approve :)

With https://github.com/processing/processing4/commit/4673b6b1f8f2ed3bb995879d2be0b0d486406267 reapplied as well, this lets nix build latest Processing again.

catilac:
Thank you @neobrain i'm going to change the settings, i need to update it to be less....rude

neobrain:
@catilac Awesome, cheers!

Just to make sure, are you going to re-apply https://github.com/processing/processing4/commit/4673b6b1f8f2ed3bb995879d2be0b0d486406267 separately as well? :)

btw completely unrelated note, but given Gradle 9 is supported now -- should the nix package update to that already or should we stick to Gradle 8 for now? How about JDK 17?

catilac:
@neobrain Oo! I should re apply that one. Thanks for pointing it out!
Yes, we are now on Gradle 9. And for now remaining on JDK 17.

We have a flag -PenableWebGPU that when set to true puts us on JDK 25 atm, too. But that is much later down the line

PR #1555, opened 2026-07-27
Follow up to unbreaking neobrains work

catilac:
Un-breaks work @neobrain did here: https://github.com/processing/processing4/commit/4673b6b1f8f2ed3bb995879d2be0b0d486406267

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -369,12 +369,12 @@ tasks.register<Zip>("zipDistributable"){
     dependsOn("createDistributable", "setExecutablePermissions")
     group = "compose desktop"
 
-    val dir = distributable().destinationDir.get()
-    val packageName = distributable().packageName.get()
+    val dir = provider { distributable().destinationDir.get() }
+    val packageName = provider { distributable().packageName.get() }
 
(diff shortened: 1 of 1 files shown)


catilac:
@neobrain could you review this? I had to change the syntax a bit, but let me know.

neobrain:
Hey, thanks for the follow up -- why did this need to be patched in signApp instead of zipDistributable? As-is I'm still getting the "Task with name 'createDistributable' not found in project 'app'" error since the `zipDistributable` task is eagerly using `distributable().`.

neobrain:
@catilac Do you need any further information? I believe my original patch should still apply without changes.

neobrain (changes requested):
See comment

catilac:
Didn't mean to be alarming @neobrain I added you as a reviewer to see that I'm able since you've been given triage permissions. Will address your comments, haven't had a chance as of yet

catilac:
@neobrain oopsie! just moved it and re-requesting review. thanks for all your help on this.

neobrain (approved):
This is working now, thanks!

PR #1557, opened 2026-07-31
fix #1520

### LINKED ISSUE #1520: `set()` method throws `NullPointerException` for `PGraphicsOpenGL` objects

Core/Environment/Rendering

### Processing version

4.5.2

### Operating system

Linux (Unbuntu) and MacOS Tahoe

### Bug description

I get a null pointer exception when I pass a `PGraphics` object to the `background()` method.

For some reason the PDE doesn't display the stack trace, I just see this:

```bash
NullPointerException
NullPointerException
```

### Steps to reproduce this

The example code below reproduces the bug.

### snippet

```processing
PGraphics g;

void setup() {
  size(300, 300, P3D);
  
  g = createGraphics(width, height, P2D);
  g.beginDraw();
  g.background(0, 255, 0);
  g.endDraw();
}

void draw() {
  background(g);
}
```

I am able to get a stack trace when I use the equivalent code in py5. I traced the problem to the `PGraphics` object's null `pixels` array.

One possible workaround is to insert a `g.loadPixels()` call after `g.endDraw()`. That fixes the problem. Another workaround is to `get()` a `PImage` object:

```processing
PGraphics g;
PImage img;

void setup() {
  size(300, 300, P3D);
  
  g = createGraphics(width, height, P2D);
  g.beginDraw();
  g.background(0, 255, 0);
  g.endDraw();
  img = g.get();
}

void draw() {
  background(img);
}
```

### Additional context

Either workaround suggests a solution. I'm happy to make a PR to fix this, you can assign it to me.

Yes, I’d like to help with this

hx2A:
Fix the `set()` method when passing a `PImage` that is a `PGraphicsOpenGL` instance. This ensures that the pixels are loaded and there is no null pointer exception.

### DIFF
--- a/core/src/processing/opengl/PGraphicsOpenGL.java
+++ b/core/src/processing/opengl/PGraphicsOpenGL.java
@@ -5512,8 +5512,11 @@ public class PGraphicsOpenGL extends PGraphics {
                          int targetX, int targetY) {
     updatePixelSize();
 
-    // Copies the pixels
+    // make sure pixel arrays are available
     loadPixels();
+    sourceImage.loadPixels();
+
+    // Copies the pixels
     int sourceOffset = sourceY * sourceImage.pixelWidth + sourceX;
     int targetOffset = targetY * pixelWidth + targetX;
     for (int y = sourceY; y < sourceY + sourceHeight; y++) {


catilac:
Amazing! Thanks so much @hx2A

catilac (approved):

hx2A:
> Amazing! Thanks so much @hx2A

You are welcome!

Issue #1522, opened 2026-06-02
RFC: future PDE implementation as a VSCode fork


kjhollen:
### Most relevant area for this enhancement?

PDE

### Feature to enhance

Maintainability

### Enhancement details

I’ve been curious about using another open source code editor as the core user interface for the Processing Development Environment (PDE). Some high level goals for such an implementation would be:

- Ease our maintenance burdens by leveraging a larger open source project 
- Use a newer front-end UI toolkit to access a larger pool of potential contributors
- And most importantly, continue to provide a UI that is friendly to beginners and artists who don’t want to write their own build scripts, as the PDE has done for nearly 25 years!

The part of the VSCode source released under the MIT License is the core of several other programming environments, including Cursor and Theia. There are some important features we can leverage from the VSCode codebase: developer extensions, syntax highlighting for many different programming languages, internationalization, telemetry, and AI code support. VSCode’s user interface is built with Typescript and Electron, which are both widely used in front-end web development. VSCode is also a project maintained by a well-resourced team of engineers and designers at Microsoft, who can put a lot more time into maintenance and security than our small (but mighty) team from Processing Foundation.

<img width="1440" height="900" alt="Image" src="https://github.com/user-attachments/assets/5a6d51ea-e6b9-4614-b0b1-3460d4163745" />

I used Copilot/Claude to build a PDE-like UI from a fork of the VSCode project for evaluative purposes, based on a suggestion from @tychedelia . This proof of concept incorporates some of the PDE’s most important beginner-friendly features in the VSCode UI, such as the welcome screen and a toolbar with run & stop buttons. It uses the Processing VSCode Extension for sketch execution. The code is visible in [this repository & branch](https://github.com/kjhollen/vscode-pde-prototype/tree/proof-of-concept). This document provides an overview of that prototype and enumerates unresolved issues for consideration.

The prototype is not being presented for code review, as it is not aligned with the processing4 repo’s AI policy. Creating this prototype was a useful exercise for me to understand what parts of the VSCode repository might be most relevant for creating a PDE-like experience. This documentation contains my notes and what I learned from reading/reviewing the prototype implementation.

I welcome community feedback on this approach, suggestions for other open source editors to evaluate as an alternative to VSCode, and comparisons to other possible PDE evolutions from other Processing Foundation projects (including processing4, the p5.js Web Editor, and the prototype collaborative editor).

## VSCode Prototype Implementation
[...]

PR #1209, opened 2025-08-15
migrating Preferences class into stand-alone version

AhmedMagedC:
## Changes
- breaks the `Preferences` class into two classes: `app.utils.Preferences` and `app.Preferences`
- `app.utils.Preferences` class is the stand-alone version making the loading and saving of preferences independent
- `app.Preferences` depends on `app.utils.Preferences` for the main functionalties (loading, saving, getting a value) but also adds other `app` dependent functionalties (fonts, language, ...)
- gracefully handles exceptions by adding a centralized custom Exception: `PreferencesException`
- loading of `default.txt` is now via JAR resource system

## Todo
- [ ] the SettingFolder location depends on the `Base` class, which forces you to explicitly set the location when using `app.utils.Preferences`. We can decouple it from `Base` class making it independt of `app` module
- [ ] Adding callback events/listeners when a preference changes
- [ ] Tests that will declare and test the functionality of Preferences
- [ ] A Github Action that will run the Tests on all supported platforms

PR #1563, opened 2026-08-06
Make Preferences standalone in app:utils

### LINKED ISSUE #1104: Making `Preferences` standalone

diyaayay:
## Resolves #1104

Splits `processing.app.Preferences` into a standalone storage class (`processing.utils.Preferences` in `app:utils`) and an app-side facade, so preferences work in contexts without the PDE: the Gradle plugin, CLI tools, and the preprocessor, as motivated in #1104 and the standalone-CLI discussion in #1522.

## Changes

- **New `processing.utils.Preferences`**: the storage half of the old class. Bundled defaults from the classpath, reading/writing `preferences.txt`, platform-specific keys, and the 3.x sketchbook migration. JDK-only (no `core`, no AWT/Swing). Errors are reported by **throwing `IOException` with the cause chained** instead of showing dialogs, so callers decide how to react.
- **`processing.app.Preferences` becomes a facade**: keeps fonts/colors, `Messages` dialogs, and the proxy/native-chooser side effects in their original startup order, delegating storage to the utils class. **No call sites change** (~246 across `app` and `java`).
- **`defaults.txt` moves into `app:utils` resources**: the published utils jar now carries its own defaults and works standalone. All readers (including the Compose `PreferencesProvider`) load it via `getResourceAsStream`, so nothing else changes.
- **`Platform.getName()`**: platform name for preference-key suffixes, mirroring `PConstants.platformNames`. The four strings are part of the preferences file format, so a small mirror avoids giving `app:utils` a `core` dependency.
- **Change listeners** (from the issue's TODO list): `addChangeListener` / `removeChangeListener` on the utils class. Fires only on real changes from `set()`/`unset()`; bulk loading is silent; a throwing listener can't block the change or other listeners. No consumers wired yet. First candidates are PDE components that currently re-read on window close, and possibly converging with the Compose side's `PreferencesEvents` later.
- **Preprocessor shim now delegates to the same class**: `java:preprocessor`'s `processing.app.Preferences` used to re-read `preferences.txt` on every call with the platform charset and had no defaults, so a standalone preprocessor could crash on `getInteger("editor.tabs.size")` with no prefs file present. It now resolves from the bundled defaults. *Happy to split this commit into a follow-up PR if preferred.*

## Design notes

- **Why the cyclic dependency from the earlier attempt is gone:** the original blocker was that the settings-folder logic lived in `:app` (`Base`/`Platform`) while `Preferences` was moving to `:app:utils`, so each module needed the other. Since then, #1196 created `app:utils` and #1335 moved settings-folder resolution into `processing.utils.Settings`, so the knowledge `Preferences` needs now lives below it and all dependency arrows point one way. This PR adds no new dependencies on top of that: the remaining app couplings are handled by throwing instead of dialogs, and by **passing the settings folder in as an argument** (`loadUserPrefs(File)`), with `Settings.getFolder()` and the `processing.app.preferences.file` system property as fallbacks for standalone use.
- **Staged init:** the app needs to set its own defaults (`run.window.bgcolor`, CJK input-method support) between loading the bundled defaults and the user's file, so init is split into `loadDefaults()` / `loadUserPrefs(File)`, with `init(File)` composing them for standalone consumers. This keeps the PDE's startup order exactly as on `main`.
- **Delegation instead of inheritance:** static methods only shadow, they don't override. An `extends` relationship would let calls through the parent type silently skip the app's error dialogs. The facade delegates explicitly; the cost is fully-qualified names (Java can't alias same-named imports), which was preferred over renaming a class and churning every reference.
- **Intentional behavior changes** (both make failures visible rather than silent): a corrupt/unreadable `preferences.txt` surfaces as an exception (the PDE shows the same error dialog as before), and the utils class won't run the sketchbook migration/save over a file it couldn't read.

Builds on the direction @AhmedMagedC explored in #1209, thank you for the groundwork! Rebuilt on today's `main` since that branch predates the `app:utils` module: delegation instead of static inheritance, chained `IOException`s instead of a custom exception, no `core` dependency, and defaults bundled in the utils jar.

## Tests

- **17 new tests** in `app:utils` (`PreferencesTest`): default loading, file creation, and backwards compatibility (3.x sketchbook migration, backslash normalization, platform-suffix keys), plus save/load round trips, the `preferences.file` override, `getInteger` fallback, sorted save, and the listener contract.
- **All suites pass on the rebased branch**: `app:utils`, `app` (incl. `PreferencesKtTest`, so the Compose provider still finds the moved `defaults.txt`), and `java` (incl. `ParserTests` through the preprocessor).
- **Smoke-tested the utils jar standalone** (jshell, nothing else on the classpath): fresh folder → defaults resolve → set/save → new JVM reads the value back. This caught a first-save folder-creation bug only real standalone use could hit, fixed with a regression test.
- **Manual:** PDE runs normally; preference changes persist across restart.
## Checklist

- [x] Tests pass locally


app/src/processing/app/Preferences.java
```diff
@@ -165,120 +120,24 @@ static void handleProxy(String protocol, String hostProp, String portProp) {
...
-        String value = line.substring(equals + 1).trim();
-        if (!isPlatformSpecific(key, value, platformSpecific)) {
-          table.put(key, value);
-        }
-      }
-    }
-    // Now override the keys with any platform-specific defaults we've found.
-    for (String key : platformSpecific.keySet()) {
-      table.put(key, platformSpecific.get(key));
-    }
```
catilac, line 205: was this moved somewhere? or just deleted?

app/src/processing/app/Preferences.java
```diff
@@ -289,44 +148,31 @@ static public void save() {
...
-      (ignoring case) will just be false.. so if malformed -> false
-    if (value == null) return defaultValue;
-
-    try {
-      return (new Boolean(value)).booleanValue();
-    } catch (NumberFormatException e) {
-      System.err.println("expecting an integer: " + attribute + " = " + value);
-    }
-    return defaultValue;
-    */
```
catilac, line 329: I need help understanding why this was removed

catilac (commented):
left some questions! sorry it's taken so long to get to this. 
this is a VERY big patch and it will affect a lot of users if it goes out so we need to be very careful. we will need more folks to test and review.

